PR #11, opened 2020-02-04
Add '.xcodesamplecode.plist' file to achieve proper markdown rendering

Sajjon:
Same contents as my [merged PR into swift-numerics](https://github.com/apple/swift-numerics/pull/34), giving proper markdown rendering inside Xcode.

### Motivation:

Reading of markdown becomes easier. Extra important in Swift packages, where the `README` is the first file opened by Xcode when opening said package.

### Modifications:

Just adding the `.xcodesamplecode.plist` file (to the root).

### Result:

As seen in [Apple's ARKit demo (inside `/ARKitExample.xcodeproj/` directory)](https://developer.apple.com/sample-code/wwdc/2017/PlacingObjects.zip)

### Before:
before

### After:
after

### DIFF
--- a/.xcodesamplecode.plist
+++ b/.xcodesamplecode.plist
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<array/>
+</plist>


Lukasa:

Sajjon:
@Lukasa Thanks for merging!

### ⚠️ Editing is now disabled in Xcode
I forgot to say that this effectively disables editing markdown files from Xcode - or at least I have not found a way of doing so. @Lukasa if you know any way of temporarily allowing for editing from Xcode please share here.

The "workaround" is just to open markdown files with your favourite alternative text editor, e.g. emacs, VIM, VI, Sublime Text or whatever,

Lukasa:
I think that the limitation above is reasonable. 👍

PR #13, opened 2020-02-04
Adding `gyb` discrepancy check to `sanity` script.

Sajjon:
Adding `gyb` "discrepancy check" to `sanity` script - meaning if running `gyb` results in (git) changes, then the `sanity` scripts flags this as an error, achiveving two things
1. help to prevent users from manually editing Swift files generated by `gyb`
2. help to remind users to run gyb and git commit generated files (at a later stage we might wanna add some git hook too?)

### Motivation:
For  beginners who might not be familiar with metaprogramming. especially `gyb`, it ought to be clear that some files should not be edited directly.  Thus I think it prudent to mention this for contributors in the README. Also how to actually make use of `gyb` themselves.

### Modifications:
1) Adding `gyb` to scripts …
2) Adding warning in `gyb` files about gyb overwritting manual edits to gyb generated swift files
3) Updated README with `gyb` instructions
4) Adding script for running all `gyb` generation (`scripts/generate_boilerplate_files_with_gyb.sh`)
5) Adding CI check for discrepancies in `gyb` generated files

### Result:
Hopefully it is more clear what `gyb` and how to  run it. And how it is used, and which files that should not be modified (manually), marked with `// MARK: - Generated file, do NOT edit`.


Sajjon:
I do understand if it is highly unwanted to append the suffix `_generated_file_do_not_edit` to generated Swift files. But I think it is very important to highlight it as clearly as possible.

So an alternative approach might be to use folders? That in combination with the `// MARK: - Generated file, do NOT edit` might be enough? By using folders I mean moving
`Sources/Crypto/Signatures/ECDSA.swift` to `Sources/Crypto/Signatures/ECDSA/GeneratedDoNotEdit/ECDSA.swift`
and correspondig `.swift.gyb` file `Sources/Crypto/Signatures/ECDSA.swift.gyb` to `Sources/Crypto/Signatures/ECDSA/EditThis/ECDSA.swift.gyb` or similar...

Sources/Crypto/Digests/Digests_generated_file_do_not_edit.swift
```diff
@@ -15,6 +15,13 @@
 @_exported import CryptoKit
 #else
 
+// MARK: - Generated file, do NOT edit
+// any edits of this file WILL be overwritten and thus discarded
+// see section `GYB` in `README` for more details.
+
```
Sajjon: I've no idea why gyb adds these three additional new lines... how can I rectify this? I'm using:
```bash
gyb --line-directive ""
```

Sources/Crypto/Digests/Digests_generated_file_do_not_edit.swift
```diff
@@ -57,10 +64,13 @@ public struct SHA256Digest: DigestPrivate {
     }
     
     public func hash(into hasher: inout Hasher) {
-        self.withUnsafeBytes { hasher.combine(bytes: $0) }
+        self.withUnsafeBytes {hasher.combine(bytes: $0)}
     }
 }
 
+
+
```
Sajjon: How to remove these three newlines?

Sources/Crypto/Digests/Digests_generated_file_do_not_edit.swift
```diff
@@ -105,10 +115,13 @@ public struct SHA384Digest: DigestPrivate {
     }
     
     public func hash(into hasher: inout Hasher) {
-        self.withUnsafeBytes { hasher.combine(bytes: $0) }
+        self.withUnsafeBytes {hasher.combine(bytes: $0)}
     }
 }
 
+
```
Sajjon: How to remove these three newlines?

Sajjon:
@Lukasa You and I must have been using different versions of GYB. In the editted README of this PR you see which version I use.

I had to edit the .gyb files to get the same generated files, e.g. had to edit some whitespace. But I've not managed to fix `extension Insecure{` (no white space before `{`) generated by Digests.swift.gyb.

Should I install a different version of GYB? In root of the [`utils` directory](https://github.com/apple/swift/tree/master/utils) I can see:
1. File [gyb_stdlib_support.py](https://github.com/apple/swift/blob/master/utils/gyb_stdlib_support.py)
2. File [gyb_foundation_support.py](https://github.com/apple/swift/blob/master/utils/gyb_foundation_support.py)
3. Folder [gyb_sourcekit_support](https://github.com/apple/swift/tree/master/utils/gyb_sourcekit_support)
4. Folder [gyb_syntax_support](https://github.com/apple/swift/tree/master/utils/gyb_syntax_support)

I have not downloaded/used any of these...

Maybe it would be prudent to bundle this project with gyb? Making sure everyone is running the same version? Or is that a terrible idea? Can gyb be included in a different manner so that one can lock its version, and git include that lock, kind of like `Gemfile.lock` or how `Package.resolved` works.

Sources/Crypto/Digests/Digests_generated_file_do_not_edit.swift
Lukasa: The three new lines are coming from the gaps in between the formatting directives. I suspect a previous run of swift-format cleared them up. Try deleting the empty lines between the directives and re-running.

Lukasa: That is, the empty lines between the three sections of:

```
%{
 digests_and_length = [{"name": "SHA256", "count": 32},{"name": "SHA384","count":48},{"name":"SHA512", "count": 64},{"name":"SHA1", "count":20, "prefix":"Insecure"},{"name":"MD5", "count":16, "prefix":"Insecure"}]
 }%

% for HF in digests_and_length:
%{
    name = HF["name"]
    byteCount = HF["count"]
    wordsCount = (byteCount*8)/64 + (0 if ((byteCount*8)%64) == 0 else 1)
}%

%{
    if "prefix" in HF.keys():
        protocol_prefix = "extension " + HF["prefix"] + "{"
        protocol_suffix = "}"
    else:
        protocol_prefix = ""
        protocol_suffix = ""
}%
```

Lukasa:
I don't think we have different versions of gyb in use, just an inconsistency in re-running it: that's on me. We should add a run of gyb to the CI system and have the CI fail if gyb changes the code.

If you add a section [to the dockerfile](https://github.com/apple/swift-crypto/blob/5000bef96781cb31b003a39424722525e2a4d76d/docker/Dockerfile#L30) that downloads gyb, you can the update the [sanity script](https://github.com/apple/swift-crypto/blob/master/scripts/sanity.sh) to run gyb and fail the build if gyb changes files. You can see an equivalent usage [here with `swiftformat`](https://github.com/apple/swift-crypto/blob/5000bef96781cb31b003a39424722525e2a4d76d/scripts/sanity.sh#L24-L43).

That will let us enforce that these things stay in step in future, which will make our lives a lot easier.

Lukasa:
Broadly I'm really happy with this direction. However, I'd like to avoid the rename of the files: I think the warning banner + CI enforcement is sufficient to avoid any problems here. Can you revert that part of the change as well?

Sajjon:
> Broadly I'm really happy with this direction. However, I'd like to avoid the rename of the files: I think the warning banner + CI enforcement is sufficient to avoid any problems here. Can you revert that part of the change as well?

Understood! Will do, I will fix in an hour or so! Thanks!

Lukasa:
No rush, take your time. :grin:

Sajjon:
@Lukasa what do you think about git hooks?

If we could identify generated files somehow in a git hook (maybe `pre-commit`), we could make sure people cannot even commit changes to generated file. One way of doing that is using [the folder idea I wrote about above](https://github.com/apple/swift-crypto/pull/13#issuecomment-581896991) maybe?

Another idea, instead of using git hooks, is to lock editing? Cocoapods does this, locking all files for generated projects, I think it is [this PR which implemented that](https://github.com/CocoaPods/CocoaPods/pull/3438). This might be even better than the git hook? Since as soon as someone tries to edit the file, a warning will be displayed. Protecting developers from the scenario of doing edits to the generated file, and not until `git add` (is that when a `pre-commit` hook is triggered?).
 
Any solution should of course "work well" with CI tools...

Lukasa:
I think git hooks are useful tools and I'd be happy to add the outline of one to this repository for users who want to use it, but naturally the CI must continue to enforce the requirement so it's strictly additive to that work.

I don't love the locked editing idea: I think it's just a bit excessive, and prevents people from doing quick hacky prototypes for no good reason. I'm ok with the robot preventing things from going wrong in the master branch.

tomerd:

Lukasa:
@Sajjon To be clear, I think I'd like to keep the autogenerated files with their names and paths as they were: just with a comment at the top of them.

Sajjon:
@Lukasa Got it, I was just playing around with a `pre-commit`  hook. 

It is really too bad that one cannot change an open PR to a **draft** on Github.

Please ignore this PR until I ping you, when I've made changes in Dockerfile and sanity script :)

Lukasa:
Heh, 👍 will do!

Sajjon:
@Lukasa Last jigsaw in the puzzle is to be able to call `gyb` from the santity script. The usage of the `swiftformat` invocation works because the Dockerfile places it in `$HOME/.tools`, but I might not wanna put `gyb` there on my local machine. And I might wanna be able to use `gyb`  locally too.

I did not know any simple way of "bulk" gyb, so created this simple bash script `generate_boilerplate_files_with_gyb.sh`:

```bash
FULL_PATH_TO_GYB=$1
find . -name '*.gyb' |                                               \
    while read file; do                                              \
        "${FULL_PATH_TO_GYB}" --line-directive '' -o "${file%.gyb}" "$file"; \
    done
```

Which could be simplified without any input args, if we where to **git include** `gyb` in the repo. It would also remove the need for downloading it in the dockerfile, and also simplify the sanity script (could just run `generate_boilerplate_files_with_gyb.sh` and do the git comparision).

Another advantage of putting `gyb` under version control is that we know everyone is using same version of `gyb`.

What say you? :)

Lukasa:
I'm happy to add `gyb.py` to the scripts directory.

Sajjon:
@Lukasa OK please review :) (sorry for spamming you... 😜 and thanks for the patience!)

Lukasa:

Sajjon:
hmm unclear what failed? `Build step 'Execute shell' marked build as failure`

Lukasa:
Indeed, it's a bit unclear. I'm going to run the docker-compose commands locally and see what happened: we might be missing something in the execution environment in CI.

Sajjon:
@Lukasa I think I got it right: https://github.com/apple/swift-crypto/pull/13/commits/e4699c08f182315bf9ceaa2eb000dee689202550

Lukasa:

Sajjon:
💃

PR #15, opened 2020-02-04
More idiomatic use of XCTest

ZevEisenberg:
More idiomatic use of XCTest to get better error messages if a test fails

### Motivation:

I was browsing the source code after reading the Swift blog post, and noticed that the use of XCTest could be better. For example, `XCTAssertEqual(foo, bar)` will give a better error message on failure than `XCTAssert(foo == bar)`.

### Modifications:

Make the following transformations:

| Old | New |
|-------------------------|--------------------------------|
| `XCTAssert(foo == bar)` | `XCTAssertEqual(foo, bar)` |
| `XCTAssert(foo < bar)` | `XCTAssertLessThan(foo, bar)` |
| `XCTAssert(false)` | `XCTFail()` |
| `XCTAssert(true)` | deleted, since this is a no-op |

Also cleaned up a confusing comment.

### Result:

Any tests that fail will give more useful error messages.

### DIFF
--- a/Tests/CryptoTests/Authenticated Encryption/AES-GCM-Runner.swift
+++ b/Tests/CryptoTests/Authenticated Encryption/AES-GCM-Runner.swift
@@ -70,8 +70,8 @@ class AESGCMTests: XCTestCase {
         let recoveredPlaintext = try! AES.GCM.open(ciphertext, using: key, authenticating: Data())
         let recoveredPlaintextWithoutAAD = try! AES.GCM.open(ciphertext, using: key)
 
-        XCTAssert(recoveredPlaintext == plaintext)
-        XCTAssert(recoveredPlaintextWithoutAAD == plaintext)
+        XCTAssertEqual(recoveredPlaintext, plaintext)
+        XCTAssertEqual(recoveredPlaintextWithoutAAD, plaintext)
     }
 
     func testExtractingBytesFromNonce() throws {
@@ -155,7 +155,7 @@ class AESGCMTests: XCTestCase {
             let ciphertext = try! AES.GCM.seal(message, using: key, nonce: nonce, authenticating: aad)
             let recoveredPlaintext = try! AES.GCM.open(ciphertext, using: key, authenticating: aad)
 
-            XCTAssert(Array(recoveredPlaintext) == Array(message))
+            XCTAssertEqual(Array(recoveredPlaintext), Array(message))
         }
 
         let message = Array("Hello, world, it's AES-GCM!".utf8)
@@ -189,7 +189,7 @@ class AESGCMTests: XCTestCase {
                                 do {
                                     nonce = try AES.GCM.Nonce(data: nonceData)
                                 } catch {
-                                    XCTAssert(nonceData.count < 12)
+                                    XCTAssertLessThan(nonceData.count, 12)
                                     continue
                                 }
 
@@ -211,22 +211,22 @@ class AESGCMTests: XCTestCase {
 
                                 let sb = try! AES.GCM.seal(msg, using: key, nonce: nonce, authenticating: aad)
 
-                                XCTAssert(Data(ct) == sb.ciphertext)
+                                XCTAssertEqual(Data(ct), sb.ciphertext)
 
                                 if (testVector.result == "valid") {
-                                    XCTAssert(Data(tag) == sb.tag)
+                                    XCTAssertEqual(Data(tag), sb.tag)
                                 }
 
                                 do {
                                     let recovered_pt = try AES.GCM.open(AES.GCM.SealedBox(nonce: nonce, ciphertext: ct, tag: tag), using: key, authenticating: aad)
 
                                     if (testVector.result == "valid" || testVector.result == "acceptable") {
-                                        XCTAssert(recovered_pt == msg)
+                                        XCTAssertEqual(recovered_pt, msg)
                                     } else {
-                                        XCTAssert(false)
+                                        XCTFail()
                                     }
                                 } catch {
-                                    XCTAssert(testVector.result == "invalid")
+                                    XCTAssertEqual(testVector.result, "invalid")
                                 }
                             } catch {
                                 XCTAssert(testVector.result == "invalid" || testVector.iv == "")
--- a/Tests/CryptoTests/Authenticated Encryption/AES-GCM-Runner.swift
+++ b/Tests/CryptoTests/Authenticated Encryption/AES-GCM-Runner.swift
@@ -148,14 +148,14 @@ class AESGCMTests: XCTestCase {
     }
 
     func testRoundTripDataProtocols() throws {
-        func roundTrip<Message: DataProtocol, AAD: DataProtocol>(message: Message, aad: AAD) {
+        func roundTrip<Message: DataProtocol, AAD: DataProtocol>(message: Message, aad: AAD, file: StaticString = #file, line: UInt = #line) {
             let key = SymmetricKey(size: .bits256)
             let nonce = AES.GCM.Nonce()
 
             let ciphertext = try! AES.GCM.seal(message, using: key, nonce: nonce, authenticating: aad)
             let recoveredPlaintext = try! AES.GCM.open(ciphertext, using: key, authenticating: aad)
 
-            XCTAssertEqual(Array(recoveredPlaintext), Array(message))
+            XCTAssertEqual(Array(recoveredPlaintext), Array(message), file: file, line: line)
         }
 
         let message = Array("Hello, world, it's AES-GCM!".utf8)
--- a/Tests/CryptoTests/Authenticated Encryption/AES-GCM-Runner.swift
+++ b/Tests/CryptoTests/Authenticated Encryption/AES-GCM-Runner.swift
@@ -41,7 +41,7 @@ struct AESGCMTestVector: Codable {
 }
 
 class AESGCMTests: XCTestCase {
-    func testingBadKeySize() {
+    func testBadKeySize() {
         let plaintext: Data = "Some Super Secret Message".data(using: String.Encoding.utf8)!
         let key = SymmetricKey(size: .init(bitCount: 304))
         let nonce = AES.GCM.Nonce()
--- a/Tests/CryptoTests/Authenticated Encryption/AES-GCM-Runner.swift
+++ b/Tests/CryptoTests/Authenticated Encryption/AES-GCM-Runner.swift
@@ -53,22 +53,22 @@ class AESGCMTests: XCTestCase {
         let ciphertext = Array("This is some weird ciphertext".utf8)
         let tag = Array(repeating: UInt8(0), count: 16)
 
-        let regularNonce = try AES.GCM.Nonce(data: Array(repeating: 0, count: 12))
-        let longNonce = try AES.GCM.Nonce(data: Array(repeating: 0, count: 13))
+        let regularNonce = try orFail { try AES.GCM.Nonce(data: Array(repeating: 0, count: 12)) }
+        let longNonce = try orFail { try AES.GCM.Nonce(data: Array(repeating: 0, count: 13)) }
 
         XCTAssertNotNil(try AES.GCM.SealedBox(nonce: regularNonce, ciphertext: ciphertext, tag: tag).combined)
         XCTAssertNil(try AES.GCM.SealedBox(nonce: longNonce, ciphertext: ciphertext, tag: tag).combined)
     }
 
-    func testEncryptDecrypt() {
+    func testEncryptDecrypt() throws {
         let plaintext: Data = "Some Super Secret Message".data(using: String.Encoding.utf8)!
 
         let key = SymmetricKey(size: .bits256)
         let nonce = AES.GCM.Nonce()
 
-        let ciphertext = try! AES.GCM.seal(plaintext, using: key, nonce: nonce)
-        let recoveredPlaintext = try! AES.GCM.open(ciphertext, using: key, authenticating: Data())
-        let recoveredPlaintextWithoutAAD = try! AES.GCM.open(ciphertext, using: key)
+        let ciphertext = try orFail { try AES.GCM.seal(plaintext, using: key, nonce: nonce) }
+        let recoveredPlaintext = try orFail { try AES.GCM.open(ciphertext, using: key, authenticating: Data()) }
+        let recoveredPlaintextWithoutAAD = try orFail { try AES.GCM.open(ciphertext, using: key) }
 
         XCTAssertEqual(recoveredPlaintext, plaintext)
         XCTAssertEqual(recoveredPlaintextWithoutAAD, plaintext)
@@ -80,8 +80,8 @@ class AESGCMTests: XCTestCase {
 
         let testNonceBytes = Array(UInt8(0)..<UInt8(12))
         let (contiguousNonceBytes, discontiguousNonceBytes) = testNonceBytes.asDataProtocols()
-        let nonceFromContiguous = try AES.GCM.Nonce(data: contiguousNonceBytes)
-        let nonceFromDiscontiguous = try AES.GCM.Nonce(data: discontiguousNonceBytes)
+        let nonceFromContiguous = try orFail { try AES.GCM.Nonce(data: contiguousNonceBytes) }
+        let nonceFromDiscontiguous = try orFail { try AES.GCM.Nonce(data: discontiguousNonceBytes) }
 
         XCTAssertEqual(Array(nonceFromContiguous), testNonceBytes)
         XCTAssertEqual(Array(nonceFromDiscontiguous), testNonceBytes)
@@ -98,8 +98,8 @@ class AESGCMTests: XCTestCase {
         let ciphertext = Array("This pretty clearly isn't ciphertext, but sure why not".utf8)
         let (contiguousCiphertext, discontiguousCiphertext) = ciphertext.asDataProtocols()
 
-        let contiguousSB = try AES.GCM.SealedBox(combined: contiguousCiphertext)
-        let discontiguousSB = try AES.GCM.SealedBox(combined: discontiguousCiphertext)
+        let contiguousSB = try orFail { try AES.GCM.SealedBox(combined: contiguousCiphertext) }
+        let discontiguousSB = try orFail { try AES.GCM.SealedBox(combined: discontiguousCiphertext) }
         XCTAssertEqual(contiguousSB.combined, discontiguousSB.combined)
         XCTAssertEqual(Array(contiguousSB.nonce), Array(discontiguousSB.nonce))
         XCTAssertEqual(contiguousSB.ciphertext, discontiguousSB.ciphertext)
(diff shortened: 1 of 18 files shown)


Tests/CryptoTests/Authenticated Encryption/ChaChaPoly-Runner.swift
```diff
@@ -189,22 +189,20 @@ class ChaChaPolyTests: XCTestCase {
...
+                XCTAssertEqual(Data(tag), sb.tag)
             }
 
             do {
                 let recovered_pt = try ChaChaPoly.open(ChaChaPoly.SealedBox(nonce: nonce, ciphertext: ct, tag: tag), using: key, authenticating: aad)
 
                 if (testVector.result == "valid" || testVector.result == "acceptable") {
-                    XCTAssert(recovered_pt == msg)
-                } else {
-                    XCTAssert(true)
```
ZevEisenberg: `XCTAssert(true)` is a no-op, so I deleted it.

Tests/CryptoTests/Signatures/EdDSA/EdDSATests.swift
```diff
@@ -72,7 +72,7 @@ class EdDSATests: XCTestCase {
 
         XCTAssert(privateKey.publicKey.isValidSignature(signature, for: DispatchData.empty))
 
-        // This should fail
+        // This signature should be invalid
```
ZevEisenberg: Clarified comment.

Sajjon:
Nice! Not relating to your changes, but while looking through the file `AES-GCM-Runner` you made changes in, I saw `testingBadKeySize`. This is the only function which does not use standard `func test<NAME>` syntax. Wanna change it it `testBadKeySize` ?

Lukasa:

Tests/CryptoTests/Authenticated Encryption/AES-GCM-Runner.swift
```diff
@@ -155,7 +155,7 @@ class AESGCMTests: XCTestCase {
             let ciphertext = try! AES.GCM.seal(message, using: key, nonce: nonce, authenticating: aad)
             let recoveredPlaintext = try! AES.GCM.open(ciphertext, using: key, authenticating: aad)
 
-            XCTAssert(Array(recoveredPlaintext) == Array(message))
+            XCTAssertEqual(Array(recoveredPlaintext), Array(message))
```
Sajjon: Not relating to your changes, but while making improvments: We should make use of `line` argument of `XCTAssertFOO` functions when using *nested calls* within tests.

So the test `testRoundTripDataProtocols` which makes use of nested function `roundTrip` can be improved from:

```swift
 func testRoundTripDataProtocols() throws {
        func roundTrip<Message: DataProtocol, AAD: DataProtocol>(message: Message, aad: AAD) {
            let key = SymmetricKey(size: .bits256)
            let nonce = AES.GCM.Nonce()

            let ciphertext = try! AES.GCM.seal(message, using: key, nonce: nonce, authenticating: aad)
            let recoveredPlaintext = try! AES.GCM.open(ciphertext, using: key, authenticating: aad)

            XCTAssertEqual(Array(recoveredPlaintext), Array(message))
        }

        let message = Array("Hello, world, it's AES-GCM!".utf8)
        let aad = Array("I heard you like Counter Mode, so I put a Galois on it".utf8)
        let (contiguousMessage, discontiguousMessage) = message.asDataProtocols()
        let (contiguousAad, discontiguousAad) = aad.asDataProtocols()

        roundTrip(message: contiguousMessage, aad: contiguousAad)
        roundTrip(message: discontiguousMessage, aad: contiguousAad)
        roundTrip(message: contiguousMessage, aad: discontiguousAad)
        roundTrip(message: discontiguousMessage, aad: discontiguousAad)
    }
```

to:

```swift
   func testRoundTripDataProtocols() throws {
        func roundTrip<Message: DataProtocol, AAD: DataProtocol>(message: Message, aad: AAD, line: UInt = #line) {
            let key = SymmetricKey(size: .bits256)
            let nonce = AES.GCM.Nonce()

            let ciphertext = try! AES.GCM.seal(message, using: key, nonce: nonce, authenticating: aad)
            let recoveredPlaintext = try! AES.GCM.open(ciphertext, using: key, authenticating: aad)

            XCTAssertEqual(Array(recoveredPlaintext), Array(message), line: line)
        }

        let message = Array("Hello, world, it's AES-GCM!".utf8)
        let aad = Array("I heard you like Counter Mode, so I put a Galois on it".utf8)
        let (contiguousMessage, discontiguousMessage) = message.asDataProtocols()
        let (contiguousAad, discontiguousAad) = aad.asDataProtocols()

        roundTrip(message: contiguousMessage, aad: contiguousAad)
        roundTrip(message: discontiguousMessage, aad: contiguousAad)
        roundTrip(message: contiguousMessage, aad: discontiguousAad)
        roundTrip(message: discontiguousMessage, aad: discontiguousAad)
    }
```

Observe the appending of `, line: line)` at end of `XCTAssertEqual`, this ensures that failing test marker in Xcode propages to the call site of each `roundTrip` call (lines: 166-169)

Sajjon: I have not checked other nested function calls in tests yet, just found this now. If calling a helper function (from a `test***` function) that are in a different file one should pass `file: file` as well as `line: line`.

ZevEisenberg: ooh great call. This is [right up my alley](https://github.com/realm/SwiftLint/pull/3040) 😁

ZevEisenberg:
Found and fixed a few more

Tests/CryptoTests/Authenticated Encryption/AES-GCM-Runner.swift
```diff
@@ -70,8 +70,8 @@ class AESGCMTests: XCTestCase {
         let recoveredPlaintext = try! AES.GCM.open(ciphertext, using: key, authenticating: Data())
```
Sajjon: I think we should change the signature if this function to be throwing, and remove the `!` from `try!`. As seen in many of the `AESGCMTests` tests.

Tests/CryptoTests/Authenticated Encryption/ChaChaPoly-Runner.swift
```diff
@@ -66,7 +66,7 @@ class ChaChaPolyTests: XCTestCase {
         let ciphertext = try! ChaChaPoly.seal(plaintext, using: key, nonce: nonce, authenticating: Data())
```
Sajjon: I think we should change the signature if this function to be throwing, and remove the `!` from `try!`. As seen in many of the `AESGCMTests` tests.

Tests/CryptoTests/Digests/DigestsTests.swift
```diff
@@ -67,9 +67,9 @@ class DigestsTests: XCTestCase {
     }
     
     func testMD5() {
-        XCTAssert(Data(Insecure.MD5.hash(data: Data())).count == Insecure.MD5.byteCount)
-        try! XCTAssert(Data(Insecure.MD5.hash(data: Data())) == Data(hexString: "d41d8cd98f00b204e9800998ecf8427e"))
-        try! XCTAssert(Data(Insecure.MD5.hash(data: Data("abcdbcdecdefdefgefghfghighijhijkijkljklmklmnlmnomnopnopq".utf8))) == Data(hexString: "8215ef0796a20bcaaae116d3876c664a"))
+        XCTAssertEqual(Data(Insecure.MD5.hash(data: Data())).count, Insecure.MD5.byteCount)
+        try! XCTAssertEqual(Data(Insecure.MD5.hash(data: Data())), Data(hexString: "d41d8cd98f00b204e9800998ecf8427e"))
```
Sajjon: I think we should change the signature if this function to be throwing, and remove the `!` from `try!`. As seen in many of the `AESGCMTests` tests.

Sajjon:
@ZevEisenberg great work! While you are cleaning up some tests function, please review usage of `try!` in test functions (as per some comments - but missed some....) :)

Tests/CryptoTests/Authenticated Encryption/AES-GCM-Runner.swift
ZevEisenberg: +1 here's my favorite utility for that, since just marking a test function as `throws` is kind of lackluster.

```swift
/// Takes a throwing closure. Returns the result if the closure succeeds.
/// If the closure throws, this method registers an XCTest failure and rethrows the error.
/// - Note: closure comes last so that trailing closure syntax works.
func orFail<T>(file: StaticString = #file, line: UInt = #line, _ closure: () throws -> T) throws -> T {
    do {
        return try closure()
    }
    catch {
        XCTFail("Function threw error: \(error)", file: file, line: line)
        throw error
    }
}
```

Tests/CryptoTests/Authenticated Encryption/AES-GCM-Runner.swift
Sajjon: Swifty minds think a like

Tests/CryptoTests/Authenticated Encryption/AES-GCM-Runner.swift
Sajjon: agreed! I use stuff like that as well, good idea! :D

ZevEisenberg:
@Sajjon as often happens, this ended up being much bigger than anticipated once the dust had settled. Let me know if it's not what you had in mind.

Lukasa:

Lukasa:
There are some minor formatting issues here, can you fix those up?

ZevEisenberg:
@Lukasa which formatting issues are you referring to? I just ran SwiftFormat, and found a bunch of stuff. It looked like it hadn't been run in a while. Are there plans to get continuous integration going on this project, so stuff like this will get caught automatically?

tomerd:

ZevEisenberg:

Lukasa:
This project is CI’d. In this case, the CI was reporting the failures.  😉

Lukasa:

ZevEisenberg:
Yep, I see it now. It hadn't reported back yet when I commented. I think it should be good now. (Also, not sure If I'm allowed to ask the bot to test, so thanks for pinging it for me 😎.)

ZevEisenberg:
Looks like I've got some failures in the code that does `@testable import CryptoKit`, which doesn't run when I run tests on my Mac. Is there any way for me to build that variant? Otherwise, it's going to take a little bit to address all errors.

Lukasa:
Try uncommenting the commented out define in Package.swift.

ZevEisenberg:
Ah, nice. I'd missed that. Should be all set for a review now.

Lukasa:

Lukasa:
Looks like something introduced in the ECDH tests is leading to test failure. We should take a look there.

Tests/CryptoTests/ECDH/BoringSSL/secpECDH_Runner_boring.swift
```diff
@@ -25,34 +25,35 @@ import XCTest
...
+                XCTAssert(testVector.flags.contains("CompressedPoint") || testVector.result == "invalid" || testVector.flags.contains("InvalidPublic") || testVector.flags.contains("InvalidAsn"), file: file, line: line)
             } catch ECDHTestErrors.ParseSPKIFailure {
-                XCTAssert(testVector.flags.contains("InvalidAsn") || testVector.flags.contains("UnnamedCurve"))
+                XCTAssert(testVector.flags.contains("InvalidAsn") || testVector.flags.contains("UnnamedCurve"), file: file, line: line)
             } catch {
                 if testVector.result == "valid" {
-                    XCTAssert(testVector.tcId == 31 || testVector.tcId == 20 || testVector.tcId == 25)
+                    XCTAssert(testVector.tcId == 31 || testVector.tcId == 20 || testVector.tcId == 25, file: file, line: line)
                 }
             }
```
Lukasa: This test is not quite the same. Some errors were accepted here, and now they all trigger test failure.

ZevEisenberg: Oops, I had overzealously wrapped some things that were actually supposed to throw during testing. Fixed now.

Lukasa:

Tests/CryptoTests/Authenticated Encryption/AES-GCM-Runner.swift
```diff
@@ -143,96 +143,98 @@ class AESGCMTests: XCTestCase {
...
 
-                                if testVector.tag.count > 0 {
-                                    tag = try! Array(hexString: testVector.tag)
-                                }
+                            if testVector.tag.count > 0 {
+                                tag = try orFail { try Array(hexString: testVector.tag) }
+                            }
 
-                                let sb = try! AES.GCM.seal(msg, using: key, nonce: nonce, authenticating: aad)
+                            let sb = try orFail { try AES.GCM.seal(msg, using: key, nonce: nonce, authenticating: aad) }
```
Lukasa: I'm a bit nervous about the `orFail`s in this `do` block. I think we should aim to remove them, because they are handled in the `catch` below.

Tests/CryptoTests/ECDH/X25519-Runner.swift
```diff
@@ -55,34 +55,38 @@ class X25519Tests: XCTestCase {
...
-            let privateKey = try! Curve25519.KeyAgreement.PrivateKey(rawRepresentation: Array(hexString: testVector.privateKey))
+            let publicKey = try orFail { try Curve25519.KeyAgreement.PublicKey(rawRepresentation: Array(hexString: testVector.publicKey)) }
+            let privateKey = try orFail { try Curve25519.KeyAgreement.PrivateKey(rawRepresentation: Array(hexString: testVector.privateKey)) }
 
             do {
-                let expectedSharedSecret = try Array(hexString: testVector.shared)
+                let expectedSharedSecret = try orFail { try Array(hexString: testVector.shared) }
 
-                XCTAssert(try Array(privateKey.sharedSecretFromKeyAgreement(with: publicKey).ss) == expectedSharedSecret)
+                let testSharedSecret = try orFail { try Array(privateKey.sharedSecretFromKeyAgreement(with: publicKey).ss) }
```
Lukasa: These `orFail`s should also come out.

Tests/CryptoTests/Authenticated Encryption/AES-GCM-Runner.swift
ZevEisenberg: Agreed. It was a mistake. Will fix.

ZevEisenberg: I also removed a few that were inside `XCTAssert*` functions, since those are already set up to catch errors, I believe.

Sajjon:
@ZevEisenberg @Lukasa Ah just saw the [release notes of Xcode 11.4 beta](https://developer.apple.com/documentation/xcode_release_notes/xcode_11_4_beta_release_notes)

> Xcode’s New File templates for XCTest in Swift now include the throws keyword for all test methods. Errors thrown by test methods are recorded as test failures. (56619497)

Which might remove benefits of `orFail` method?

Lukasa:
As I understand it, those benefits are not present in older Xcodes or on Linux. Going forward we probably want to update `orFail` to use a check. Something like:

```swift
func orFail<T>(file: StaticString = #file, line: UInt = #line, _ closure: () throws -> T) throws -> T {
    #if os(Linux)
    do {
        return try closure()
    } catch {
        XCTFail("Function threw error: \(error)", file: file, line: line)
    }
    #else
    return try closure()
    #endif
}
```

FredericJacobs:
I think adding the conditional as @Lukasa suggests until support lands in CoreLibs would be a good solution.

ZevEisenberg:
@FredericJacobs might we want to hold off on adding the conditional until Xcode 11.4 is out of beta? Until then, people are likely to be working on this in 11.3.1, and the `#if os(Linux)` won't apply in that case.

stmontgomery:
> @FredericJacobs might we want to hold off on adding the conditional until Xcode 11.4 is out of beta? Until then, people are likely to be working on this in 11.3.1, and the `#if os(Linux)` won't apply in that case.

I think you could improve this wrapper a bit more, so that it "passes through" the thrown error if (a) the target OS is new enough to support the new Swift runtime hook for error location capture _-and-_ (b) the compiler is 5.2 or later, indicating that the Xcode is new enough (11.4 or later) to include the new XCTest logic for capturing error locations. I haven't tried this, but I'm envisioning something like

```swift
func orFail<T>(file: StaticString = #file, line: UInt = #line, _ closure: () throws -> T) throws -> T {
    func wrapper<T>(_ closure: () throws -> T, file: StaticString, line: UInt) throws -> T {
        do {
            return try closure()
        } catch {
            XCTFail("Function threw error: \(error)", file: file, line: line)
            throw error
        }
    }

    #if compiler(>=5.2) && !os(Linux)
        if #available(macOS 10.15.4, iOS 13.4, tvOS 13.4, *) {
            return try closure()
        } else {
            return try wrapper(closure, file: file, line: line)
        }
    #else
        return try wrapper(closure, file: file, line: line)
    #endif
}
```

Tests/CryptoTests/Utils/XCTestUtils.swift
```diff
@@ -0,0 +1,61 @@
...
+        do {
+            return try closure()
+        } catch {
+            XCTFail("Function threw error: \(error)", file: file, line: line)
+            throw error
+        }
+    }
+
+    #if compiler(>=5.2) && !os(Linux)
+        if #available(macOS 10.15.4, iOS 13.4, tvOS 13.4, *) {
```
Lukasa: Can we add `watchOS` and `macCatalyst` to the availability guard here please?

ZevEisenberg: Which watchOS version should I use? Testing isn't _currently_ supported on watchOS, right?

Lukasa: I'd just use the same as the rest of the project.

ZevEisenberg: Ah, great call. I grabbed watchOS 6.0 from Package.swift 👍🏻

Tests/CryptoTests/Utils/XCTestUtils.swift
```diff
@@ -0,0 +1,61 @@
...
+        do {
+            return try closure()
+        } catch {
+            XCTFail("Function threw error: \(error)", file: file, line: line)
+            throw error
+        }
+    }
+
+    #if compiler(>=5.2) && !os(Linux)
+        if #available(macOS 10.15.4, macCatalyst 10.15.4, iOS 13.4, tvOS 13.4, watchOS 6.0, *) {
```
Lukasa: macCatalyst is versioned the same as iOS, not the same as macOS.

ZevEisenberg: TIL. Fixed ☺️

Lukasa:

Tests/CryptoTests/Utils/XCTestUtils.swift
```diff
@@ -0,0 +1,61 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftCrypto open source project
+//
+// Copyright (c) 2019 Apple Inc. and the SwiftCrypto project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftCrypto project authors
```
Lukasa:
```suggestion
// See CONTRIBUTORS.md for the list of SwiftCrypto project authors
```

Tests/CryptoTests/Utils/XCTestUtils.swift
```diff
@@ -0,0 +1,61 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftCrypto open source project
+//
+// Copyright (c) 2019 Apple Inc. and the SwiftCrypto project authors
```
Lukasa:
```suggestion
// Copyright (c) 2020 Apple Inc. and the SwiftCrypto project authors
```

Lukasa:

Tests/CryptoTests/Utils/XCTestUtils.swift
```diff
@@ -0,0 +1,61 @@
...
+    func wrapper<T>(_ closure: () throws -> T, file: StaticString, line: UInt) throws -> T {
+        do {
+            return try closure()
+        } catch {
+            XCTFail("Function threw error: \(error)", file: file, line: line)
+            throw error
+        }
+    }
+
+    #if compiler(>=5.2) && !os(Linux)
```
broadwaylamb:
```suggestion
    #if compiler(>=5.2) && canImport(Darwin)
```

This would be more general and better convey the intent, since there's nothing Linux-specific here.

Lukasa:

Lukasa:
Thanks for this patch!

PR #19, opened 2020-02-05
`// MARK:` delimiter in gyb generated files to allow overview in Xcode minimap 

Sajjon:
**💡 WIP needs to decide format before merge**
Adding `// MARK:` delimiter on a different format (need to agree which format is best) for different curves within gyb generated files.

### Motivation:
Some of over gyb generated files are very long, without any `// MARK:` delimiter, making it hard to get an overview in the Xcode minimap.

### Modifications:

Just adding `// MARK:` in `gyb` files.

#### before
This is is `ECDSA.swift` before `MARK` change
before_no_mark

#### after (simple)
Same file -  `ECDSA.swift` - after, with simple `// MARK: - ${CURVE} + PrivateKey`
after_simple_curve_name_

#### after (decorated)
This is `ECDH.swift` with a more decorated variant. 

**TIL** (today I learned): You cannot put multiple `// MARK` on subsequent lines, Xcode will not render them all. E.g. 3 lines of MARK will result in Xcode only displaying the first. The "workaround" this is to put comment (non `// MARK` lines) in between the 3 `// MARK` lines, which is what I've done in this decorated variant.  Thus, in the source, the MARK takes up 9 lines, but might be worth it, because the rendered MARK in the minimap is a much more vibrant delimiter:

##### Code
```swift
// MARK: *************************
//===----------------------------------------------------------------------===//
//===----------------------------------------------------------------------===//
//===----------------------------------------------------------------------===//
// MARK: -~=*  Curve ${CURVE}   *=~-
//===----------------------------------------------------------------------===//
//===----------------------------------------------------------------------===//
//===----------------------------------------------------------------------===//
// MARK: *************************

```

##### minimap
after_decorated

### DIFF
--- a/Sources/Crypto/AEADs/Nonces.swift
+++ b/Sources/Crypto/AEADs/Nonces.swift
@@ -19,6 +19,7 @@ import Foundation
 // any edits of this file WILL be overwritten and thus discarded
 // see section `gyb` in `README` for details.
 
+// MARK: - AES.GCM + Nonce
 extension AES.GCM {
     public struct Nonce: ContiguousBytes, Sequence {
         let bytes: Data
@@ -48,6 +49,7 @@ extension AES.GCM {
     }
 }
 
+// MARK: - ChaChaPoly + Nonce
 extension ChaChaPoly {
     public struct Nonce: ContiguousBytes, Sequence {
         let bytes: Data
--- a/Sources/Crypto/AEADs/Nonces.swift.gyb
+++ b/Sources/Crypto/AEADs/Nonces.swift.gyb
@@ -28,6 +28,7 @@ nonceSize = cipher["recommendedNonceSize"]
 nonceValidation = cipher["nonceValidation"]
 }%
 
+// MARK: - ${name} + Nonce
 extension ${name} {
     public struct Nonce: ContiguousBytes, Sequence {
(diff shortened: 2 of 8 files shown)


Sajjon:
@Lukasa Please let me know:
1) If you think this is a good idea at all (but I have a feeling you'll like it)
2) (if so:) Which format you prefer

Lukasa:
I definitely like this idea. I think I like it in its current form: the vivid delimiter is extremely vivid, and probably a bit excessive. The smaller marks, however, are very nice and likely to be helpful.

Sajjon:
Hehe yes, definitely a bit excessive :D 

Of the non-decorated marks, this is still `ECDH.swift`, which of these three versions do you like the best? Low, medium or high detail level? The only difference between medium and high is that medium omits repetition of Curve name, whereas high displays that again.

# Low
low_detail_level

# Medium
medium_detail_level

# High
high_detail_level

Lukasa:
I think low is fine: the structure of the code is sufficient to work out what's going on once you have the marks.

Sajjon:
@Lukasa ready to be merged, I've changed to the low detail level as per your comment 👍

Lukasa:

PR #28, opened 2020-02-24
Mitigate Folly/BoringSSL static linking issue.

Lukasa:
Motivation:

BoringSSL and Folly both attempt to play silly buggers with the sdallocx
symbol defined in jemalloc. Unfortunately, they don't play nicely
together, and the end result if you have both present is that BoringSSL
will reliably SIGBUS or SIGSEGV on initialization. Naturally, this isn't
good.

See also apple/swift-nio-ssl#191.

Modifications:

- Resolve the issue by patching out BoringSSL's allocator optimisation.
- Update BoringSSL to 6432bb46ab44731567ec923e6c8fc182f13d0070.

Result:

Flow and BoringSSL will play better together.


Lukasa:
@tomerd Looks like the API breakage script is busted?

Lukasa:

Lukasa:
Nevermind @tomerd, resolved it with a rebuild.

PR #30, opened 2020-02-28
update docs generation script to work better with selinux

tomerd:
motivation: docs scripts is broken on selinux due to restriction on ssh-agent

changes:

* change the jazzy workspace to a directory accessible by the CI node so we can push from the agent instead of from the docker container
* dont remove older docs when pushing new ones
* add version information and source links


tomerd:
@Lukasa @weissi

PR #34, opened 2020-04-10
Simplify development on macOS

dmbryson:
Support development on Apple platforms without editing Package.swift by
checking if the relevant setting has been set in the environment.

### Motivation:

Development of swift-crypto on Apple platforms currently requires manually editing Package.swift to set the required build setting `CRYPTO_IN_SWIFTPM_FORCE_BUILD_API`. The build setting change is undesirable in non-development scenarios.

### Modifications:

Alter Package.swift to conditionally set `CRYPTO_IN_SWIFTPM_FORCE_BUILD_API` if the environment variable of the same name is set during the build.

### Result:

Development on Apple platforms can be done without altering Package.swift.  For example:

```
$ CRYPTO_IN_SWIFTPM_FORCE_BUILD_API=1 swift build
```

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -23,12 +23,16 @@
 // BoringSSL Commit: 5298ef99bf2b2d77600b3bb74dd572027bf495be
 
 import PackageDescription
+import Foundation
 
-let swiftSettings: [SwiftSetting] = [
+var swiftSettings: [SwiftSetting] = [
     .define("CRYPTO_IN_SWIFTPM"),
-    // To develop this on Apple platforms, uncomment this define.
-    // .define("CRYPTO_IN_SWIFTPM_FORCE_BUILD_API"),
 ]
+
+// To develop this on Apple platforms, define this in your environment:
+if ProcessInfo.processInfo.environment["CRYPTO_IN_SWIFTPM_FORCE_BUILD_API"] != nil {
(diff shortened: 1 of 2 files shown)


Lukasa:
@aciidb0mb3r Hey Ankit, just a quick question: is it safe to assume that SwiftPM will respect environment variables? In particular, does SwiftPM guarantee to invalidate the build state and rebuild if the environment changes?

aciidb0mb3r:
In general, a package manifest should have no side-effects because swiftpm doesn't make any guarantee when and how often it will evaluate package manifests. In the current implementation, it does invalidate the manifest cache if the env changes but that's not officially documented and can change in the future.

dmbryson:
🤔 good to know.   I suppose I hadn't considered whether I could perhaps just directly set the build setting on the 'swift build' command line.   Would that propagate to dependencies?

aciidb0mb3r:
yah, -X* flags are global and apply to all packages

dmbryson:
And those would invalidate the build when changed/omitted?

dmbryson:
In either case, probably safer to do that, and not rely on undocumented behavior.  Closing this.

aciidb0mb3r:
> And those would invalidate the build when changed/omitted?

Yes.

Lukasa:
Thanks @aciidb0mb3r for the clarification and @dmbryson for the proposal!

PR #36, opened 2020-06-01
Update BoringSSL to 53a17f55247101105ae35767d5c5a6c311843a8e

Lukasa:
Updates the vendoring script to handle the slightly changed layout of BoringSSL and removed the workaround for Folly's weak linking dance as it's no longer needed.


scripts/vendor-boringssl.sh
```diff
@@ -188,7 +188,7 @@ PATTERNS=(
 'crypto/*/*/*.S'
 'crypto/*/*/*/*.c'
 'third_party/fiat/*.h'
-'third_party/fiat/*.c'
+#'third_party/fiat/*.c'
```
weissi, line 191: is this commented rather than deleted deliberately?

Lukasa: Yes: the presence of C files in this directory changes over time, and I got sick of deleting and re-adding this line all the time, so I just commented it out.

PR #41, opened 2020-06-27
DERTests: enable test function for Android

buttaface:
Enable single test erroring on Android

### Checklist

### Motivation:

No reason for this to error on Android

### Modifications:

Add Android to the list.

### Result:

All 153 tests pass when building natively on Android with a swift master snapshot from last month, instead of this lone test segfaulting before. Maybe it would be better if this list of systems were removed altogether, as it seems to include every system this codebase has been tested on.


Lukasa:
Yeah, this is mostly just a weirdly defensive guard. We should discuss removing it, but for now let’s just do this.

Lukasa:

PR #46, opened 2020-07-23
Update to Developer Beta 3 build.

Lukasa:


Lukasa:
The API failure note is irrelevant: we have not shipped the error that is being removed in this patch.

Sources/Crypto/ASN1/ECDSASignature.swift
```diff
@@ -0,0 +1,55 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.md for the list of SwiftCrypto project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+#if (os(macOS) || os(iOS) || os(watchOS) || os(tvOS)) && CRYPTO_IN_SWIFTPM && !CRYPTO_IN_SWIFTPM_FORCE_BUILD_API
+@_exported import CryptoKit
+#else
+import Foundation
```
glbrntt, line 17: What's being used from `Foundation` here?

Sources/Crypto/Signatures/ECDSA.swift
```diff
@@ -197,11 +229,15 @@ extension P384.Signing {
...
-            return self.openSSLDERRepresentation
-            #endif
+            let raw = rawRepresentation
+            let half = raw.count / 2
+            let r = Array(raw.prefix(upTo: half))[...]
+            let s = Array(raw.suffix(from: half))[...]
+
+            let sig = ASN1.ECDSASignature(r: r, s: s)
+            var serializer = ASN1.Serializer()
+            try! serializer.serialize(sig)
```
glbrntt, line 239: Why is the `!` okay here?

Lukasa: The `!` is ok for two reasons:

1. This is a `var`, so we can't throw even if we wanted to: crashing is our only accessible error handling mode
2. This is not being computed over arbitrary data but previously-validated data, so it should not be possible for this to fail. If it did, it would be our error.

Sources/Crypto/ASN1/ECDSASignature.swift
Lukasa: Nothing, but this is a pretty ordinary starting stanza and I'm loathe to change it. :wink:

Sources/Crypto/Signatures/ECDSA.swift
glbrntt: Thanks for the explainer!

PR #53, opened 2020-11-30
build: allow building on Windows

compnerd:
This allows us to build on Windows, though not entirely.  The assembly
routines need to be generated for the Windows targets, but this is
required to get to that point.  With this set of changes, we are able to
get to the point where we can start linking the `crypto-shasum` target,
which exposes the missing symbols.

_[One line description of your change]_

### Checklist

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -46,7 +46,17 @@ let package = Package(
     ],
     dependencies: [],
     targets: [
-        .target(name: "CCryptoBoringSSL"),
+        .target(
+          name: "CCryptoBoringSSL",
+          cSettings: [
+            /*
+             * This define is required on Windows, but because we need older
+             * versions of SPM, we cannot conditionally define this on Windows
+             * only.  Unconditionally define it instead.
+             */
+            .define("WIN32_LEAN_AND_MEAN"),
+          ]
+        ),
         .target(name: "CCryptoBoringSSLShims", dependencies: ["CCryptoBoringSSL"]),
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @Lukasa @tomerd

Lukasa (approved):

Lukasa:
@swift-server-bot add to whitelist

Lukasa:

Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.1
+// swift-tools-version:5.3
```
Lukasa, line 1: I'm afraid we can't bump the tools version: swift-crypto supports back to 5.1. We _can_ add a new Package.swift if we have to, though it complicates matters a bit.

Is there any particular reason not to just define WIN32_LEAN_AND_MEAN in all cases?

compnerd: The best reason that I have against that is hygiene and the reminder that it is in the user's namespace (it is not underscore prefixed, so this symbol is perfectly cromulent to be used by a user in their macOS program).  I can just unconditionally define it.  We could also sprinkle it amongst the various source files as necessary under a guard.

Lukasa: I think I'm happy to unconditionally define it for this target. cSettings are totally available in the 5.1 tools version, we just can't use _platform-specific_ settings.

compnerd: I can agree to that and dislike it right? :-p.  Done.

Lukasa (approved):

PR #56, opened 2020-12-01
CCryptoBoringSSL: add Windows generated assembly

compnerd:
This adds the generated assembly files for Windows.  Unfortunately, the
x64 generated assembly appears to be ELF-esque and SYSV ABI rather than
Win64.  Additionally, the ARM targets do not seem to generate valid
Windows ARM code either (quick perusal shows at least invalid SP
handling on ARMv7 NT and non-conforming prologues on AArch64).  Add x86
Windows assembly only for the time being.

_[One line description of your change]_

### Checklist
- [ ] I've followed the code style of the rest of the project
- [ ] I've read the [Contribution Guidelines](CONTRIBUTING.md)

### Motivation:

### Modifications:

### Result:


Lukasa:
Hmm, I'm surprised the x86_64 asm isn't right, as BoringSSL uses it on Windows for things like Chromium. Did you try asking for `nasm` instead of `coff`?

FredericJacobs (approved):

compnerd:
Using nasm would introduce a new dependency for building.  I think that after the next re-branch, might be good to try the masm variant.

Lukasa:
I'm definitely mostly interested in what we can arrange for clang to actually build, so I'm fairly relaxed about what we end up doing here.

Lukasa (approved):

PR #65, opened 2021-01-28
Update/Fix CMake build

yim-lee:
Motivation:
SwiftPM's CMake build
(https://github.com/apple/swift-package-manager/pull/3202) unsuccessful
even with https://github.com/apple/swift-crypto/pull/60.

Modifications:
- Use consistent `SWIFT_CRYPTO_EXPORTS` naming. In some places `CRYPTO_EXPORTS` was used instead.
- Install `CCryptoBoringSSL` so we can import it in code (see https://github.com/apple/swift-package-manager/pull/3202)
- Always set `BUILD_SHARED_LIBS`

Result:
Successful CMake build.


tomerd:
@compnerd ptal

compnerd:
Hmm, does the client need to link against CCryptoBoringSSL or just import it?  (that is, it needs the headers and the module map?)  Installation of just the library would be insufficient for that.  In fact, if the module is needed because it is a dependency for the rest of the code, I think that the better thing might be to use it as `@_implementationOnly`.

compnerd:
Just to make sure I understand everything - this is just a renaming of the exports file and nothing more?

yim-lee:
Not installing `CCryptBoringSSL` seems ok actually, or at least I can't get it to fail anymore, so I backed out that part of the changeset.

yim-lee:
> this is just a renaming of the exports file and nothing more

This is renaming exports file and vars plus always setting `BUILD_SHARED_LIBS` to `YES`.

compnerd (approved):
LGTM

tomerd:
@Lukasa good to merge?

Lukasa (approved):

PR #69, opened 2021-02-04
Update vendor-boringssl to save modulemap

yim-lee:
Motivation:
The vendor-boringssl script rewrites `include`, which leads to deletion of `module.modulemap`.

Modification:
Save a copy of modulemap to tmp dir and copy it back at the end.

Result:
`module.modulemap` remains after running the script.

### DIFF
--- a/scripts/vendor-boringssl.sh
+++ b/scripts/vendor-boringssl.sh
@@ -3,7 +3,7 @@
 ##
 ## This source file is part of the SwiftCrypto open source project
 ##
-## Copyright (c) 2019 Apple Inc. and the SwiftCrypto project authors
+## Copyright (c) 2019-2021 Apple Inc. and the SwiftCrypto project authors
 ## Licensed under Apache License v2.0
 ##
 ## See LICENSE.txt for license information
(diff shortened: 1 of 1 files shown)


compnerd (approved):
Thanks!

yim-lee:

scripts/vendor-boringssl.sh
```diff
@@ -145,6 +145,9 @@ if ! hash ${sed} 2>/dev/null; then
     exit 43
 fi
 
+echo "SAVING modulemap"
+cp "$DSTROOT/include/module.modulemap" "$TMPDIR"
```
Lukasa, line 149: Rather than copy this back and forth, it would be better for us to just create it wholesale every time. This is what we do with the umbrella header, and it allows this script to avoid relying on any code being present in the repository at all.

Lukasa (changes requested):

scripts/vendor-boringssl.sh
yim-lee: amended https://github.com/apple/swift-crypto/pull/69/commits/06bd913bdc8577e524d71b188889eb182a917586

Lukasa (approved):
Excellent, looks great to me. Thanks @yim-lee!

PR #80, opened 2021-04-08
Provide initial proposal for RSA signing

Lukasa:
Motivation:

Some critical use-cases require RSA signing support. This is most
commonly required when interacting with legacy systems. While RSA is an
older and less efficient cryptosystem than the elliptic-curve algorithms
supported by Swift Crypto already, those interop use-cases are
unavoidable.

This version of the patch does not promote RSA to supported API at this
time, as we'd like to let the API bake somewhat. Users are free to use
it, but these APIs are not covered by semantic versioning at this time
and may break in future.

Modifications:

- Defined new RSA API
- Implemented on top of `Security` when available
- Implemented on top of BoringSSL when `Security` is unavailable

Result:

Users can do RSA if needed.


Sources/_CryptoExtras/Util/DigestType.swift
```diff
@@ -0,0 +1,54 @@
+//===----------------------------------------------------------------------===//
```
Lukasa, line 1: This file is copied-and-updated from the core Crypto library.

Sources/_CryptoExtras/Util/Error.swift
```diff
@@ -0,0 +1,17 @@
+//===----------------------------------------------------------------------===//
```
Lukasa, line 1: This file is copied and updated from the core crypto library.

Sources/_CryptoExtras/Util/PEMDocument.swift
```diff
@@ -0,0 +1,114 @@
+//===----------------------------------------------------------------------===//
```
Lukasa, line 1: This file is copied and updated from the core Crypto library.

Tests/_CryptoExtrasTests/Utils/BytesUtil.swift
```diff
@@ -0,0 +1,101 @@
+//===----------------------------------------------------------------------===//
```
Lukasa, line 1: This is copied wholesale from the CryptoTests target.

Tests/_CryptoExtrasTests/Utils/Wycheproof.swift
```diff
@@ -0,0 +1,36 @@
+//===----------------------------------------------------------------------===//
```
Lukasa, line 1: Copied from the CryptoExtras test wholesale.

Sources/_CryptoExtras/RSA/RSA.swift
```diff
@@ -0,0 +1,311 @@
...
+    ///
+    /// The default padding is PSS using MGF1 with same hash function as produced the digest being
+    /// signed, and a salt that is as long as the digest. Note that this API will not select any
+    /// known-insecure digests.
+    ///
+    /// - Parameters:
+    ///   - signature: The signature to verify
+    ///   - digest: The digest that was signed.
+    /// - Returns: True if the signature is valid, false otherwise.
+    public func isValidSignature<D: Digest>(_ signature: _RSA.Signing.RSASignature, for digest: D) throws -> Bool {
```
weissi, line 248: I'd recommend using different (and probably more than 1 letter) generics names for `D: Digest` and `D: DataProtocol`

Sources/_CryptoExtras/RSA/RSA_security.swift
```diff
@@ -0,0 +1,320 @@
...
+        let keyData: Data
+        if let pkcs8Data = data.pkcs8RSAKeyBytes {
+            keyData = pkcs8Data
+        } else {
+            keyData = data
+        }
+
+        let key = SecKeyCreateWithData(keyData as CFData, keyAttributes as CFDictionary, &error)
+
+        if let error = error {
```
weissi, line 113: this needs double checking. Typically in the ObjC/CF APIs you are only allowed to look at the error value if you have previously determined that there was an error. Given that you `nil` initialise this, this is likely totally fine, although I believe that you're not supposed to even look at the `NSError **/CFError **` output if you don't know (from the return value) that there was an error.

Sources/_CryptoExtras/RSA/RSA_security.swift
```diff
@@ -0,0 +1,320 @@
...
+    init(keySize: _RSA.Signing.KeySize) throws {
+        let keyAttributes: [CFString: Any] = [
+            kSecAttrKeyType: kSecAttrKeyTypeRSA,
+            kSecAttrKeyClass: kSecAttrKeyClassPrivate,
+            kSecAttrKeySizeInBits: keySize.bitCount
+        ]
+        var error: Unmanaged<CFError>? = nil
+        let key = SecKeyCreateRandomKey(keyAttributes as CFDictionary, &error)
+
+        if let error = error {
```
weissi, line 133: same note here

Sources/_CryptoExtras/RSA/RSA_security.swift
```diff
@@ -0,0 +1,320 @@
...
+            throw CryptoKitError.incorrectParameterSize
+        }
+
+        self.backing = unwrappedKey
+    }
+
+    var derRepresentation: Data {
+        var error: Unmanaged<CFError>? = nil
+        let representation = SecKeyCopyExternalRepresentation(self.backing, &error)
+        precondition(error == nil)
```
weissi, line 147: this one too

Sources/_CryptoExtras/Util/BoringSSLHelpers.swift
```diff
@@ -0,0 +1,94 @@
...
+        return try block(bio)
+    }
+}
+
+extension Data {
+    init(copyingMemoryBIO bio: UnsafeMutablePointer<BIO>) throws {
+        var innerPointer: UnsafePointer<UInt8>? = nil
+        var innerLength = 0
+
+        guard 1 ==  CCryptoBoringSSL_BIO_mem_contents(bio, &innerPointer, &innerLength) else {
```
weissi, line 58:
```suggestion
        guard 1 == CCryptoBoringSSL_BIO_mem_contents(bio, &innerPointer, &innerLength) else {
```

Sources/_CryptoExtras/Util/BoringSSLHelpers.swift
```diff
@@ -0,0 +1,94 @@
...
+        self = Data(UnsafeBufferPointer(start: innerPointer, count: innerLength))
+    }
+}
+
+extension String {
+    init(copyingUTF8MemoryBIO bio: UnsafeMutablePointer<BIO>) throws {
+        var innerPointer: UnsafePointer<UInt8>? = nil
+        var innerLength = 0
+
+        guard 1 ==  CCryptoBoringSSL_BIO_mem_contents(bio, &innerPointer, &innerLength) else {
```
weissi, line 71:
```suggestion
        guard 1 == CCryptoBoringSSL_BIO_mem_contents(bio, &innerPointer, &innerLength) else {
```

Sources/_CryptoExtras/Util/DigestType.swift
```diff
@@ -0,0 +1,54 @@
...
+import Crypto
+
+struct DigestType {
+    var dispatchTable: OpaquePointer
+
+    var nid: CInt
+
+    var digestLength: Int
+
+    private init(_ dispatchTable: OpaquePointer, _ nid: CInt, digestLength: Int) {
```
weissi, line 25: mind documenting (maybe in the parameter name) who owns the OpaquePointer?

Lukasa:
Feel free to ignore all the JSON documents, they are test fixtures.

weissi (commented):
Awesome, this looks mostly good to me. Left some (minor) comments

weissi:
> Feel free to ignore all the JSON documents, they are test fixtures.

Do they have any license?

Sources/_CryptoExtras/RSA/RSA.swift
glbrntt, line 248: Is this meant to be `throws`?

Sources/_CryptoExtras/RSA/RSA.swift
```diff
@@ -0,0 +1,311 @@
...
+    /// Verifies an RSA signature with the given padding over a message with the default padding.
+    ///
+    /// SHA256 is used as the hash function. The default padding is PSS using MGF1 with SHA256
+    /// and a 32-byte salt.
+    ///
+    /// - Parameters:
+    ///   - signature: The signature to verify
+    ///   - data: The data that was signed.
+    /// - Returns: True if the signature is valid, false otherwise.
+    public func isValidSignature<D: DataProtocol>(_ signature: _RSA.Signing.RSASignature, for data: D) throws -> Bool {
```
glbrntt, line 261: Same here, is this meant to have `throws` in the signature?

Sources/_CryptoExtras/RSA/RSA_security.swift
Lukasa: Yes, this seems like good advice. I'll refactor this to check the return code first.

Sources/_CryptoExtras/RSA/RSA.swift
Lukasa: While I broadly agree, these are in line with the equivalent generic type names for the main Crypto API. I think the consistency is valuable here.

Sources/_CryptoExtras/RSA/RSA.swift
Lukasa: Nope, neither of these are, thanks.

0xTim:
I've started spiking moving JWTKit to use this in vapor/jwt-kit#55 and the basics work great and remove a load of C code 🎉 

The two things that I can't find that we use are RSA certificates (Used by Firebase) and constructing a key from JWKS parts (used by Sign In With Apple and others). It could well be me using it wrong so let me know if that's the case!

Lukasa:
There is no X.509 support in Swift Crypto at any stage, so we haven't exposed RSA certs here either. Regarding the JWK format, I assume you mean the construction of RSA from the raw numbers (i.e. `n` and `e` for public keys, `n`, `e`, `d`, and `p` plus the optional extra Chinese remainder theorem parameters for private keys)?

0xTim:
Correct, we use it [here](https://github.com/vapor/jwt-kit/blob/main/Sources/JWTKit/JWTSigners.swift#L157-L185)

Lukasa:
Noted, I'll do some circulation and see if we can work out an API we're happy with. The biggest design wrinkle there is that there isn't a natural bignum type yet, so it's not totally clear how one should design the API to ingest the larger parameters.

Sources/_CryptoExtras/RSA/RSA.swift
weissi: I see, makes sense!

Sources/_CryptoExtras/RSA/RSA_security.swift
```diff
@@ -0,0 +1,318 @@
...
+        // that it's 3 bytes long.
+        let lengthLength = Int(self[1])
+        guard lengthLength == 0x82 else {
+            return nil
+        }
+
+        let length = Int(self[2]) << 8 | Int(self[3])
+        guard length == self.count - 4 else {
+            return nil
+        }
```
glbrntt, line 295: I don't follow this:

> We'll expect it to be 3 bytes long, with the first byte telling us that it's 3 bytes long.

If the first byte, `lengthLength` is `0x82`, how is that telling us it's 3 bytes long? And then why is `length` computed from only 2 bytes?

Sources/crypto-shasum/main.swift
```diff
@@ -13,6 +13,7 @@
 //===----------------------------------------------------------------------===//
 import Foundation
 import Crypto
+import _CryptoExtras
```
glbrntt, line 16: Unused import? (Maybe same story for Package.swift too)

glbrntt (commented):
I think it'd be useful to add a note to the README or elsewhere stating that crypto extras isn't public API and is subject to change.

Other than that I left a couple of questions but looks great otherwise!

FredericJacobs (approved):

Sources/_CryptoExtras/RSA/RSA_security.swift
Lukasa: `0x82` is an ASN.1-ism. It means: "the length of this field is definite, and the length field is the two octets after this one".

glbrntt (approved):

yim-lee (approved):
Thanks @Lukasa 🙏

weissi (approved):
🎉

Lukasa:
For now I'm merging without support for RSA numbers. Still work to do to flesh out that API.

dreampiggy:
@Lukasa  Hi. I'm porting an API for RSA encrypt/decrypt from macOS/iOS to Linux.

The exist API implementation on macOS/iOS, use RSA encrypt/decrypt via `SecKeyEncrypt` and `SecKeyDecrypt` from Apple's security, where Linux not.

However, after I review the sample and test from this PR, I found there are only way matching the `SecKeyVerifySignature` and `SecKeyCreateSignature`.

And, even if I want to call the underline boringssl to do the things, I found it little hard to handle the C side for pem, pointer, etc :(

Can swift-crypto provide the same context with top level API, like this:

```swift
extension _RSA.Signing.PublicKey {
    public func encrypt<D: DataProtocol>(for data: D) throws -> DataProtocol
}

extension _RSA.Signing.PrivateKey {
    public func decrypt<D: DataProtocol>(for data: D) throws -> DataProtocol
}
```

Thanks anyway 👍

Lukasa:
That proposed API is a good example of why we thus far have avoided doing RSA encryption/decryption. When doing encryption/decryption you have to have an opinion about padding modes. PKCS#1 v1.5 padding should be avoided in favour of OAEP-based padding, but it's hard to avoid _offering_ PKCS#1 v1.5 padding as it's widely used. The downside as a result is that users have to make a choice about what padding option to use, which is extremely hard to explain coherently.

Ideally you'd avoid using RSA encryption altogether where practical.

PR #83, opened 2021-05-27
Add SECURITY.md for swift-crypto

Lukasa:
The SSWG requires that projects now prominently link to their security process. As ours is delegated to Apple, we should expand upon the existing text in our README and provide an obvious place to look for that process.


weissi (approved):
awesome, ta

PR #84, opened 2021-05-27
Futz around with Swift versions.

Lukasa:
Motivation:

The various Apple SoS projects are currently dropping Swift 5.1 support.
Swift Crypto should do the same. While we're here, let's add CI for 5.4,
5.5 nightlies, and main nightlies.

Modifications:

- Remove 5.1 dockerfile.
- Stop claiming we support 5.1.
- Add 5.4, 5.5, and main dockerfiles.
- Updated Swift Format to 5.2.
- Associated formatting changes.

Result:

Better test matrix and support matrix.

PR #85, opened 2021-05-31
Fix build for 32-bit platforms

buttaface:
Motivation:

I couldn't build for [Android armv7 on the daily CI](https://github.com/buttaface/swift-android-sdk/actions?query=event%3Aschedule) I set up [without this](https://github.com/buttaface/swift-android-sdk/blob/main/package-patches/swift-crypto-armv7.patch).

Modifications:

Pass the right size of UInt to BoringSSL.

Result:

I can build for armv7 again.

### Checklist

Really, we should just be able to pass `UInt` but the stupid Swift compiler calls that an error.


Sources/_CryptoExtras/Util/BoringSSLHelpers.swift
```diff
@@ -86,7 +86,11 @@ extension FixedWidthInteger {
             CCryptoBoringSSL_BN_clear(&bn)
         }
 
+        #if arch(arm) || arch(i386)
+        CCryptoBoringSSL_BN_set_word(&bn, UInt32(self))
+        #else
         CCryptoBoringSSL_BN_set_word(&bn, UInt64(self))
+        #endif
```
Lukasa, line 93: Thanks for this patch! Instead of the conditional compilation, can we use `.init(self)` instead and let the type inference engine figure it out?

Lukasa: Relatedly, the `precondition` on line 81 is clearly wrong in the 32-bit environment. Can we flip that over to `precondition(self.bitWidth <= UInt.bitWidth)`?

Lukasa (changes requested):
Thanks for this! A small tweak would be nice, and then I'm happy to merge this.

Sources/_CryptoExtras/Util/BoringSSLHelpers.swift
buttaface: Done, I cross-compiled this new patch for Android armv7 without a problem.

Lukasa (approved):
Ace, LGTM, thanks!

Lukasa:

Lukasa:
Ah yeah, we have to merge #84 first before this can land

PR #87, opened 2021-07-29
Support compressed keys.

Lukasa:
Motivation:

CryptoKit supports the compressed key format: we should as well.

Modifications:

Extend the constructors of public keys to tolerate the compressed
representation.

Result:

Users holding compressed keys can use them.


glbrntt (approved):
LGTM

PR #90, opened 2021-08-31
fix possible over-indexing

Cyberbeni:
If a type conforming to DataProtocol has multiple regions then we would iterate over through the whole data multiple times, over-indexing the output array.

### Checklist

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Sources/Crypto/Util/PrettyBytes.swift
+++ b/Sources/Crypto/Util/PrettyBytes.swift
@@ -42,12 +42,10 @@ extension DataProtocol {
         let ptr = UnsafeMutablePointer<UInt8>.allocate(capacity: hexLen)
         var offset = 0
         
-        self.regions.forEach { (_) in
-            for i in self {
-                ptr[Int(offset * 2)] = itoh((i >> 4) & 0xF)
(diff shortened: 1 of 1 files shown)


Cyberbeni:
I was looking for a way to pretty format hex data but decided to stay with `data.map { String(format: %02x, $0) }.joined()` as it is more readable and doesn't have a measurable performance impact for my use case. I also learned that [`String(bytesNoCopy:...)` will still copy the buffer](https://forums.swift.org/t/does-string-bytesnocopy-copy-bytes/51643/10), so if the performance of hexString is critical then [`init(unsafeUninitializedCapacity:...)`](https://developer.apple.com/documentation/swift/string/3565957-init) should be used where available.

Data (at least the cross-platform implementation) [will always have 1 region](https://github.com/apple/swift-corelibs-foundation/blob/eec4b26deee34edb7664ddd9c1222492a399d122/Sources/Foundation/Data.swift#L2180-L2183), so currently it probably doesn't affect anyone but didn't want to leave the incorrect code in case it changes in the future or someone wants to use a different type that conforms to DataProtocol.

Lukasa:

Lukasa (approved):
Great catch, thanks!

PR #91, opened 2021-09-08
Update BoringSSL to 25773430c07075a368416c3646fa4b07daf4968a

Lukasa:


glbrntt (approved):

compnerd:
CC: @tomerd
This is going to break SPM builds on non-macOS targets.  This update pulls in https://boringssl-review.googlesource.com/c/boringssl/+/48985 which makes a number of structs completely opaque typedefs, which I believe are not imported into Swift via the clang importer.

yim-lee:
Thanks @compnerd. SwiftPM is still on swift-crypto 1.x so it should be fine for now, but we will definitely need to change the code when SwiftPM upgrades to swift-crypto 2.x.

ProfFan:
Just chiming in that Arch Linux builds of Swift master is breaking due to this change. It's nontrivial to debug because it fails with a message saying
```
error: cannot find type 'X509_STORE_CTX' in scope
```
but this struct is not in the Swift source code so grepping does not help.

Feel free to ignore for now, this post is to reduce Google time for future visitors.

ProfFan:
@yim-lee  I see, thanks a lot for the info! I'll pin to 1.1.4 then.

PR #94, opened 2021-11-01
Avoid unnecessary copies of BoringSSL

Lukasa:
Motivation:

When using only the CryptoKit API on Apple platforms we have always
thunked through to the CryptoKit interface and implementation. However,
we didn't do a thorough job of preventing the BoringSSL target from
getting compiled and linked.

We can do a better job now, which will save compile times and binary
sizes in many cases.

Modifications:

- Change Package.swift to express a target specific dependency in most
  cases.
- Preserve a development mode which overrides that target specific
  dependency.
- Add the missing compile guards.

Results:

Smaller binaries and faster compiles on Apple platforms.


artemredkin (approved):

FredericJacobs (approved):
⚡️

Package.swift
```diff
@@ -24,11 +24,34 @@
 
 import PackageDescription
 
-let swiftSettings: [SwiftSetting] = [
-    .define("CRYPTO_IN_SWIFTPM"),
-    // To develop this on Apple platforms, uncomment this define.
-    // .define("CRYPTO_IN_SWIFTPM_FORCE_BUILD_API"),
-]
+// To develop this on Apple platforms, set this to true
+let development = false
```
PeterAdams-A, line 28: Should we give a way of setting this without changing the code?  Eg env var or something?

PeterAdams-A (approved):

Package.swift
Lukasa: Given that the only intention is to make this available when developing locally, structuring things like this is fine.

artemredkin:

yim-lee:
This would break SwiftPM build on Apple platforms since the C module `PackageCollectionsSigningLibc` pulls in `CCryptoBoringSSL` through `Crypto`. I am looking to see if SwiftPM can define `PackageCollectionsSigningLibc` conditionally.

@compnerd Does this impact the CMake build at all?

compnerd:
> This would break SwiftPM build on Apple platforms since the C module `PackageCollectionsSigningLibc` pulls in `CCryptoBoringSSL` through `Crypto`. I am looking to see if SwiftPM can define `PackageCollectionsSigningLibc` conditionally.
> 
> 
> 
> @compnerd Does this impact the CMake build at all?

Not sure, it seems like it's just excluding some files from each target?  If so, I think it might be okay.

Lukasa:
No, we're changing the SwiftPM manifest to only build some targets on Linux.

However, I don't think this change propagates through to the CMake situation so it only impacts us on Darwin when using SwiftPM and trying to reach directly into the C targets. I think SwiftPM might be the only thing doing that.

Issue #59, opened 2020-12-16
Password based KDF


jmjauer:
#### Motivation:

Currently no password based KDF is available. This is required for tasks like generating a AES key from a password.

#### Importance:

Without it, generating crypto keys from a password is not possible.

PR #98, opened 2021-12-09
New API Proposal: PBKDF2

admkopec:
I've proposed an implementation of **PBKDF2** algorithm using the CryptoKit style. As per issue #59 

### Checklist

I haven't made any changes to `gyb` files.

### New API Proposal: Password based KDF #59 

Motivation:

I've added a proposed implementation of **PBKDF2**, as per proposal in issue #59 to allow easy creation of `SymmetricKey` from relatively low entropy data sets, such as user provided passwords.

Importance:

This API change is attempting to simplify the creation of SymmetricKeys from user entered passwords. 
It uses an ageing algorithm – **PBKDF2**, but it is still widely supported and popular. A newer algorithm – **scrypt** is added as a better, more secure alternative.

### Modifications:

I've proposed a new API:  `KDF` namespace with `Scrypt` struct and an `Insecure` namespace with `PBKDF2` struct with associated hash function and a single static method which takes a passphrase and salt, the size of the resulting key in bytes and optionally the number of rounds to use in the **PBKDF2** algorithm. 
In the case of **scrypt** algorithm there is no associated hash function due to the nature of the algorithm, the single static method takes passphrase and salt, the size of the resulting key in bytes and optionally: the number of rounds, block size and parallelism factor.
I implemented this functions using both CommonCrypto and BoringSSL where available.
The function names and parameters are heavily influenced by current HKDF implementation.

Additionally, I've added test cases to test the proposed API. I've decided to use test vectors described in RFC6070 for PBKDF2 and test vectors described in RFC7914 for scrypt.

### Result:

This will add a new API to `_CryptoKitExtras`.
It follows the naming conventions from CryptoKit's HKDF implementation.


Lukasa:
Thanks for filing this! I wanted to let you know that I've seen it, and I'm going to discuss with my colleagues to work out whether PBKDF2 is a good fit here.

admkopec:
Possibly Argon2 might fit CryptoKit better, but due to its lack of support in both CommonCrypto and BoringSSL, I decided to implement PBKDF2 instead. Please let me know when you decide, if PBKDF2 should be added to CryptoKit.

Sources/Crypto/Key Derivation/BoringSSL/PBKDF2_commoncrypto.swift
```diff
@@ -0,0 +1,76 @@
...
+            }
+        }
+        if derivationStatus != kCCSuccess {
+            throw CryptoKitError.underlyingCoreCryptoError(error: derivationStatus)
+        }
+        return SymmetricKey(data: derivedKeyData)
+    }
+}
+
+#endif
```
sbhuiyan20, line 76:
```suggestion
#endif
```

MPLew-is:
@Lukasa any other updates or thoughts on this? This is the one thing stopping me from switching fully to `swift-crypto` from other third-party packages.

nerzh:
any updates ?

Lukasa:
Not at this time, I'm afraid.

lovetodream:
There is a discussion about this going on [here](https://forums.swift.org/t/pitch-oraclenio-oracle-db-driver-built-on-swiftnio/69088). @Lukasa, how do you want to continue? Maybe @admkopec has time to move this to `_CryptoExtas`? I'm open to help, both in this case and in adding alternatives, if there's anything I can do.

Lukasa:
Sorry for the delay here. Yes, I'm open to taking a PR for PKBDF2 with the following notes:

1. The API surface should be added in `_CryptoExtras`.
2. PBKDF2 should be namespaced under a `KDF` namespace
3. PBKDF2 should be further namespaced under an `Insecure` namespace within `KDF`, to discourage its use
4. A version of `scrypt` should be added as well

Is that acceptable?

MPLew-is:
Would again advocate for an Argon2 implementation, even if it’s just wrapping [the reference C implementation](https://github.com/P-H-C/phc-winner-argon2) for now.

admkopec:
I'll start moving the implementation over to `_CryptoExtras` and will take a look at `scrypt`. But as for the better alternative, I'd still vote for `Argon2` over `scrypt` to be honest even if it would mean adding an external dependency to `_CryptoExtras`.

MPLew-is:
> even if it would mean adding an external dependency to `_CryptoExtras`

While I'm sure that there are more performant implementations, the reference implementation hasn't been updated in 3 years and the algorithm is final at this point, so you could probably even just copy the source directly into `_CryptoExtras` and avoid the external dependency. I've been running a project with basically that exact setup for a few years with no issues, running on both Mac and Linux.

Lukasa:
I'm open to us adding Argon2, though I want to have a discussion with my colleagues about whether the reference implementation is the right way to do that. But for now let's focus on PBKDF2 and scrypt: one problem at a time!

nerzh:
Perfect answer, please implement already, we've been waiting for over 2 years. We are not comfortable including openssl header in our code. Just do it please, we'd rather start writing code on Rust than wait another 2 years for Argon2

admkopec:
@Lukasa I've implemented the changes you suggested. Please verify if this is an acceptable solution.

Sources/_CryptoExtras/Key Derivation/Scrypt/BoringSSL/Scrypt_boring.swift
```diff
@@ -0,0 +1,63 @@
...
+    ///    - blockSize: The block size to be used by the algorithm.
+    ///    - parallelism: The parallelism factor indicating how many threads should be run in parallel.
+    /// - Returns: The derived symmetric key.
+    static func deriveKey<Passphrase: DataProtocol, Salt: DataProtocol>(from password: Passphrase, salt: Salt, outputByteCount: Int, rounds: Int, blockSize: Int, parallelism: Int) throws -> SymmetricKey {
+        // This should be SecureBytes, but we can't use that here.
+        var derivedKeyData = Data(count: outputByteCount)
+        let derivedCount = derivedKeyData.count
+        
+        // This computes the maximum amount of memory that will be used by the scrypt algorithm with an additional memory page to spare. This value will be used by the BoringSSL as the memory limit for the algorithm.
+        let maxMemory = 128 * rounds * blockSize * parallelism + 4096
```
Lukasa, line 41: I think it might be worth making this an optional parameter that defaults to `nil`, for which we can then compute this default. I can see value in letting users do that.

Sources/_CryptoExtras/Key Derivation/Scrypt/BoringSSL/Scrypt_boring.swift
```diff
@@ -0,0 +1,63 @@
...
+        
+        let result = derivedKeyData.withUnsafeMutableBytes { derivedKeyBytes -> Int32 in
+            let keyBuffer: UnsafeMutablePointer<UInt8> =
+            derivedKeyBytes.baseAddress!.assumingMemoryBound(to: UInt8.self)
+            let saltBytes: ContiguousBytes = salt.regions.count == 1 ? salt.regions.first! : Array(salt)
+            return saltBytes.withUnsafeBytes { saltBytes -> Int32 in
+                let saltBuffer = saltBytes.baseAddress!.assumingMemoryBound(to: UInt8.self)
```
Lukasa, line 48: Many of the notes I made in the PBKDF2 file apply here.

Sources/_CryptoExtras/Key Derivation/Scrypt/Scrypt.swift
```diff
@@ -0,0 +1,40 @@
...
+        ///
+        /// - Parameters:
+        ///    - password: The passphrase, which should be used as a basis for the key. This can be any type that conforms to `DataProtocol`, like `Data` or an array of `UInt8` instances.
+        ///    - salt: The salt to use for key derivation.
+        ///    - outputByteCount: The length in bytes of resulting symmetric key.
+        ///    - rounds: The number of rounds which should be used to perform key derivation. Must be a power of 2 less than `2^(128 * blockSize / 8)`.
+        ///    - blockSize: The block size to use for key derivation.
+        ///    - parallelism: The parallelism factor to use for key derivation. Must be a positive integer less than or equal to `((2^32 - 1) * 32) / (128 * blockSize)`.
+        /// - Returns: The derived symmetric key.
+        public static func deriveKey<Passphrase: DataProtocol, Salt: DataProtocol>(from password: Passphrase, salt: Salt, outputByteCount: Int, rounds: Int = 16_384, blockSize: Int = 8, parallelism: Int = 1) throws -> SymmetricKey {
```
Lukasa, line 36: I'm inclined _not_ to default these parameters here. We don't want to make default choices for users as they obscure the security relevance of the parameters.

Lukasa (commented):
Thanks for this! Some initial notes in the diff. I've mostly restrained my review to the product code for now, I'll review the tests later.

nerzh:
Hi @admkopec, can you please tell me if this pullrequest will be developed further ?

admkopec:
@nerzh Yes, sorry for the delay, a more urgent thing came up on my side. I'll implement the necessary changes as soon as I get a chance to get back to this PR.

nerzh:
@admkopec glad to hear you have plans to complete this much needed PR. Thanks for the quick response.

Sources/_CryptoExtras/Key Derivation/PBKDF2/PBKDF2.swift
```diff
@@ -0,0 +1,42 @@
...
+
+#if canImport(CommonCrypto)
+fileprivate typealias BackingPBKDF2 = CommonCryptoPBKDF2
+#else
+fileprivate typealias BackingPBKDF2 = BoringSSLPBKDF2
+#endif
+
+extension KDF.Insecure {
+    /// An implementation of PBKDF2 key derivation function.
+    public struct PBKDF2<H: HashFunction> {
```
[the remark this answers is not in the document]
admkopec: Sorry it took so long, is there any reason for using `public struct` over `public enum` in this situation? I have currently provided a `public enum` for this purpose.

Sources/_CryptoExtras/Key Derivation/Scrypt/Scrypt.swift
admkopec: Fair point, I have made the same assumption with PBKDF2's round for which I also removed the default value now.

admkopec:
Once again I am very sorry for the delay. I have implemented suggestions from the review. @Lukasa, Please take a look if it is acceptable now.

amosavian:
There are two other KDFs - HKDF and X9.63- which are implemented in `SharedSecret` extension. Shouldn't we move these two into the new namespace? Or we may implement PBKDF2 and Scrypt in `SharedSecret` to preserve API consistency? @Lukasa @admkopec

nerzh:
@Lukasa ping for this PR 🙂

Sources/_CryptoExtras/Key Derivation/PBKDF2/PBKDF2.swift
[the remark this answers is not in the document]
Lukasa: Yeah, `public struct` lets us add more cases without breaking API.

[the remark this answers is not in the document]
admkopec: I converted it to `public struct`, in library specific code, due to conversion into either CommonCrypto's or BoringSSL's hash functions. I added a `preconditionFailure` as the default case, as I believe we should not choose any default hash function when we are supplied an unsupported value, in order not to confuse the users.

Lukasa:

admkopec:

Sources/_CryptoExtras/Key Derivation/PBKDF2/PBKDF2.swift
```diff
@@ -0,0 +1,58 @@
...
+        public static func deriveKey<Passphrase: DataProtocol, Salt: DataProtocol>(from password: Passphrase, salt: Salt, using hashFunction: HashFunction, outputByteCount: Int, rounds: Int) throws -> SymmetricKey {
+            return try BackingPBKDF2.deriveKey(from: password, salt: salt, using: hashFunction, outputByteCount: outputByteCount, rounds: rounds)
+        }
+        
+        public struct HashFunction: Equatable, Hashable {
+            public let rawValue: String
+            
+            public static let insecure_md5 = HashFunction(rawValue: "insecure_md5")
+            public static let insecure_sha1 = HashFunction(rawValue: "insecure_sha1")
+            public static let insecure_sha224 = HashFunction(rawValue: "insecure_sha224")
```
Lukasa, line 48: NIt: let's do `insecureSHA224` and friends, instead of the underscore.

Lukasa:
Just a note that I’m on vacation right now so I can’t review, but will review when I’m back.

nerzh:
> Just a note that I’m on vacation right now so I can’t review, but will review when I’m back.

Hello, perhaps you have already rested and can take a closer look at this pull request 🙃

nerzh:
@admkopec thank you for your work and patience, I hope this pull request will be accepted before the whole world switches to argon2 or scrypt

Sources/_CryptoExtras/Key Derivation/PBKDF2/PBKDF2.swift
```diff
@@ -0,0 +1,58 @@
...
+        /// Derives a symmetric key using the PBKDF2 algorithm.
+        ///
+        /// - Parameters:
+        ///    - password: The passphrase, which should be used as a basis for the key. This can be any type that conforms to `DataProtocol`, like `Data` or an array of `UInt8` instances.
+        ///    - salt: The salt to use for key derivation.
+        ///    - hashFunction: The hash function to use for key derivation.
+        ///    - outputByteCount: The length in bytes of resulting symmetric key.
+        ///    - rounds: The number of rounds which should be used to perform key derivation.
+        /// - Returns: The derived symmetric key.
+        public static func deriveKey<Passphrase: DataProtocol, Salt: DataProtocol>(from password: Passphrase, salt: Salt, using hashFunction: HashFunction, outputByteCount: Int, rounds: Int) throws -> SymmetricKey {
```
Lukasa, line 39: This API generally looks good, but I'm wondering if we should add a minimum round count to avoid folks naively using the number `1`. We should _certainly_ add documentation suggesting a high round count (in the hundreds of thousands), but I can't see any world where we'd want to accept less than 1k.

If we have use-cases that absolutely need lower round counts I'd love to know, but those might be best served the way we serve RSA key sizes: with a second API that is a bit harder to justify typing.

Sources/_CryptoExtras/Key Derivation/PBKDF2/PBKDF2.swift
```diff
@@ -0,0 +1,58 @@
...
+            return try BackingPBKDF2.deriveKey(from: password, salt: salt, using: hashFunction, outputByteCount: outputByteCount, rounds: rounds)
+        }
+        
+        public struct HashFunction: Equatable, Hashable {
+            public let rawValue: String
```
Lukasa, line 44: I propose making this internal, we don't need to expose it to others.

Sources/_CryptoExtras/Key Derivation/PBKDF2/PBKDF2.swift
```diff
@@ -0,0 +1,58 @@
...
+        ///    - salt: The salt to use for key derivation.
+        ///    - hashFunction: The hash function to use for key derivation.
+        ///    - outputByteCount: The length in bytes of resulting symmetric key.
+        ///    - rounds: The number of rounds which should be used to perform key derivation.
+        /// - Returns: The derived symmetric key.
+        public static func deriveKey<Passphrase: DataProtocol, Salt: DataProtocol>(from password: Passphrase, salt: Salt, using hashFunction: HashFunction, outputByteCount: Int, rounds: Int) throws -> SymmetricKey {
+            return try BackingPBKDF2.deriveKey(from: password, salt: salt, using: hashFunction, outputByteCount: outputByteCount, rounds: rounds)
+        }
+        
+        public struct HashFunction: Equatable, Hashable {
```
Lukasa, line 43:
```suggestion
        public struct HashFunction: Equatable, Hashable, Sendable {
```

Sources/_CryptoExtras/Key Derivation/PBKDF2/PBKDF2.swift
```diff
@@ -0,0 +1,58 @@
...
+
+#if canImport(CommonCrypto)
+fileprivate typealias BackingPBKDF2 = CommonCryptoPBKDF2
+#else
+fileprivate typealias BackingPBKDF2 = BoringSSLPBKDF2
+#endif
+
+extension KDF.Insecure {
+    /// An implementation of PBKDF2 key derivation function.
+    public struct PBKDF2 {
```
Lukasa, line 29:
```suggestion
    public struct PBKDF2: Sendable {
```

Sources/_CryptoExtras/Key Derivation/KDF.swift
```diff
@@ -0,0 +1,29 @@
...
+//===----------------------------------------------------------------------===//
+import Crypto
+#if canImport(Darwin) || swift(>=5.9.1)
+import Foundation
+#else
+@preconcurrency import Foundation
+#endif
+
+/// A container for Key Detivation Function algorithms.
+public enum KDF {
```
Lukasa, line 22:
```suggestion
public enum KDF: Sendable {
```

Sources/_CryptoExtras/Key Derivation/KDF.swift
```diff
@@ -0,0 +1,29 @@
...
+    /// A container for older, cryptographically insecure algorithms.
+    ///
+    /// - Important: These algorithms aren’t considered cryptographically secure,
+    /// but the framework provides them for backward compatibility with older
+    /// services that require them. For new services, avoid these algorithms.
+    public enum Insecure {}
```
Lukasa, line 28:
```suggestion
    public enum Insecure: Sendable {}
```

Lukasa:
Apologies for the delays getting to this, this is now fairly high up my priority list so we should be able to get it in fairly soon.

Lukasa:

Sources/_CryptoExtras/Key Derivation/PBKDF2/PBKDF2.swift
lovetodream: I do actually have such a use case. User+password authentication in Oracle typically encrypts the password with 4096 iterations. Hashes it alongside other data using SHA512, does a bit of CBC de- and encryption with it and some more data. **And finally encrypts the end result using another 3 rounds of PBKDF2**. (https://github.com/lovetodream/oracle-nio/blob/7fee78535f68c9f3a451beeed15483bc6ac2877b/Sources/OracleNIO/Messages/Coding/OracleFrontendMessageEncoder.swift#L926-L974)

It might not be the same for every version of Oracle because this is controlled by the db server, but I confirmed it is the default behaviour with Oracle 23ai (newest release).

Lukasa: Ace, that's helpful. So I think that this then suggests we want two APIs, one like the one here and one closer that maybe uses the label `unsafeUncheckedRounds` (or similar). 

My desire is to _enable_ the use-case you have, while strongly discouraging users who pick up PBKDF2 without much thought from putting in extraordinarily low values.

lovetodream: Sounds good to me 👍

admkopec: I added the `unsafeUncheckedRounds` API method and a `guard` check in the regular API method to throw an incorrect parameter size error whenever the rounds is less than 1000. Please verify if this is acceptable. 

If we now have the unsafe API method, is the 1000 rounds requirement enough though, as it was proposed way back in original RFC with intent to increase number of rounds over time?

Lukasa: Probably not. OWASP suggests a minimum round count of 210k for SHA512-based PBKDF2, and so maybe we should consider that the lower bound.

Lukasa:
There are a few very minor tweaks to the license headers, as shown in the failing GH workflow. Otherwise this is basically good to go!

Lukasa:
Actually, let's not bother with that: you've done more than enough to get this patch over the line. I've pushed a fix for those.

Lukasa (approved):
I'm happy with this, thanks so much for your diligent work!

PR #105, opened 2022-02-25
build: attempt to workaround linker issues on Darwin

compnerd:
The use of a non-cctools librarian causes ld64 some heartburn.
Explicitly opt into using the cctools librarian.

_[One line description of your change]_

### Checklist
- [ ] I've run tests to see all new and existing tests pass
- [ ] I've followed the code style of the rest of the project
- [ ] I've read the [Contribution Guidelines](CONTRIBUTING.md)
- [ ] I've updated the documentation if necessary

#### If you've made changes to `gyb` files
- [ ] I've run `.script/generate_boilerplate_files_with_gyb` and included updated generated files in a commit of this pull request

### Motivation:

_[Explain here the context, and why you're making that change. What is the problem you're trying to solve.]_

### Modifications:

_[Describe the modifications you've done.]_

### Result:

_[After your change, what will change.]_

PR #106, opened 2022-02-25
build: attempt to workaround linker issues on Darwin (#105)

Lukasa:
The use of a non-cctools librarian causes ld64 some heartburn.
Explicitly opt into using the cctools librarian.

rdar://87850031
(cherry picked from commit facc82fbb465e8e308ceeeeb2bdd2830ae8023b3)


Lukasa:
Released as 1.1.7.

yim-lee:
Thanks @Lukasa!

PR #114, opened 2022-04-20
Use local CCryptoBoringSSL_arm_arch.h ARM header

buttaface:
Motivation:

Fix the build on these AArch64 platforms.

Modifications:

Use the local renamed header.

Result:

The linux assembly files build on Android AArch64 again and all tests pass.

### Checklist
- [ ] I've updated the documentation if necessary - no need

The recent BoringSSL update in #113 broke building this package on Android AArch64 and likely iOS too, so I looked and figured you meant to change this header.


Lukasa (changes requested):
Thanks for this patch! We don't directly make changes to the BoringSSL code: instead, we try to make changes in the form of the `vendor-boringssl.sh` script so that we don't lose this change the next time we update the copy of BoringSSL. In this case we need to take a look at why the rename didn't work correctly. In this case it seems to be because those few headers were imported using quotes instead of angle brackets. I'll follow-up with a patch to fix this up, thanks!

Lukasa:
Superseded by #115.

PR #122, opened 2022-06-21
Crypto: explicitly type the array slice

compnerd:
Without the explicit typing, the build fails with recent toolchain
builds expecting the type to conform to an incorrect type.  This
repairs the build.


compnerd:
I think that you are correct in that this is likely a regression in the toolchain, though this does repair the build.  It is likely a good idea to see if we can bisect it and understand what changed in the type deduction.

Lukasa:
I mean, FWIW this code is unchanged in 2 years so there's no doubt this is a regression in the toolchain. Additionally, Crypto is in the source code compatibility suite at a version that post-dates the introduction of this line of code, so the compat suite will fail even if we repair the build. I'm inclined to suggest that we should fix the toolchain rather than merge this change: do you particularly strongly disagree?

PR #125, opened 2022-07-21
Add API providing basic RSA pubkey encrypt and privkey decrypt

### LINKED ISSUE #124: Support for basic RSA encrypt/decrypt operation
#### Motivation:

While it is always desirable to encourage motion away from outdated and security-questionable algorithms like RSA, in practical usage it is not always possible to avoid requiring them. In the case of a MySQL protocol client, support for the `caching_sha2_password` auth plugin (the default in MySQL 8.0) on a non-TLS connection (often desired for speed in secure environments such as a LAN) requires the ability to perform an RSA encrypt operation using a public key provided by the MySQL server. (I propose including the corresponding RSA "decrypt with private key" operation as a matter of symmetry.)

#### Importance:

It is not possible to implement this functionality with existing API short of embedding an entire copy of BoringSSL in the protocol client package, which negates the usefulness of the `swift-crypto` package outright (to say nothing of the cost in duplicated code and build times).

A working implementation is available [here](https://github.com/apple/swift-crypto/pull/125).

gwynne:
Add API to `_CryptoExtras` to provide basic RSA pubkey encrypt and privkey decrypt operations.

Closes #124.

### Checklist

### Motivation:

See #124 for details.

### Modifications:

Added two new APIs to the `RSA` support under a new `RSA.Encryption` namespace, with associated data types necessary for representing the appropriate inputs and outputs. Added new test vectors and associated test cases exercising the new API surface.

Note: The extent of the exposed API has been deliberately limited to the bare minimum, as it is not desirable to expose more of RSA's core primitives than absolutely necessary.


Lukasa:

Lukasa:
Thanks for this PR @gwynne. For now I'm going to circulate this with some colleagues and get back to you.

Higher08:
There is no option to set algorithm (only SHA1 is available). It will be good to add this parameter.
Both secyrity and BoringSSL supports that.
For setting alg in BoringSSL we can use
```
EVP_PKEY_CTX_ctrl_str(ctx, "rsa_oaep_md", "sha256");
EVP_PKEY_CTX_ctrl_str(ctx, "rsa_mgf1_md", "sha256");
```

t089:
Is this something that could be merged any time soon? I've just stumbled upon this while trying to sign a JWT for a GitHub app that I could use RSA support from Crypto.

t089:
> RSA signing is already merged.

🙈 Thanks for the pointer! Looked for it in the wrong place.

0xTim:
Is there any update on this getting merged?

Lukasa:
Currently, no. It remains a source of real tension including a PKCS1v1.5 decrypt operation, and certainly as implemented here we wouldn't want to merge it, because it's not possible to use it safely (no constant-time padding check and no appropriate mitigation path). Marking the PKCS1v1.5 padding insecure is helping, but less than ideal, especially as the API does not provide a mechanism by which users could hold it safely. If we can find a way to not needing it at all, that would be supremely helpful to getting this merged.

Lukasa:
OAEP does not have a safety concern to the same degree. However, I'm open to renegotiating the question of marking the PKCS 1v1.5 padding insecure in this context if @gwynne is willing to update the PR.

gwynne:
> OAEP does not have a safety concern to the same degree. However, I'm open to renegotiating the question of marking the PKCS 1v1.5 padding insecure in this context if @gwynne is willing to update the PR.

I'm willing to just remove support for the PKCS #1 v1.5 padding altogether. It's not needed for my purposes in any event. I'll rebase the PR against the current tip of trunk and take the insecure padding out.

gwynne:
@Lukasa Fully rebased, PKCS1 v1.5 padding support removed. Also fixed two deprecation warnings in as minimal a fashion as I could manage.

Sources/_CryptoExtras/RSA/RSA.swift
```diff
@@ -318,6 +319,65 @@ extension _RSA.Signing {
     }
 }
 
+extension _RSA.Encryption {
+    public typealias PublicKey = _RSA.Signing.PublicKey
+    public typealias PrivateKey = _RSA.Signing.PrivateKey
```
Lukasa, line 324: Typealiases are just a bit too weak for use in this case. Can I suggest we add new types that just so happen to hold the Signing key types?

Sources/_CryptoExtras/RSA/RSA.swift
```diff
@@ -318,6 +319,65 @@ extension _RSA.Signing {
...
+}
+
+extension _RSA.Encryption {
+    public struct RSAEncryptedData: ContiguousBytes {
```
Lukasa, line 328: Any reason not to just use `Data` directly here?

Sources/_CryptoExtras/RSA/RSA_boring.swift
```diff
@@ -233,6 +241,33 @@ extension BoringSSLRSAPublicKey {
                 return rc == 1
             }
         }
+        
+        fileprivate func encrypt<D: DataProtocol>(_ data: D, padding: _RSA.Encryption.Padding) throws -> _RSA.Encryption.RSAEncryptedData {
+            let outputSize = Int(CCryptoBoringSSL_RSA_size(self.pointer))
+
+            let output = try Array<UInt8>(unsafeUninitializedCapacity: outputSize) { bufferPtr, length in
```
Lukasa, line 248: Pivoting through `Array` doesn't have the impact we want here, as we have to immediately allocate a `Data`. It'd be better to zero-initialize the `Data` and then use `withUnsafeMutableBytes`.

Sources/_CryptoExtras/RSA/RSA.swift
```diff
@@ -318,6 +319,65 @@ extension _RSA.Signing {
...
+        private init(_ backing: Backing) {
+            self.backing = backing
+        }
+        
+        public static let PKCS1_OAEP = Self(.pkcs1_oaep)
+    }
+}
+
+extension _RSA.Encryption.PrivateKey {
+    public func decrypt<D: DataProtocol>(_ data: D, padding: _RSA.Encryption.Padding) throws -> _RSA.Encryption.RSADecryptedData {
```
Lukasa, line 370: As constructed, these encrypt/decrypt functions require messages to be of limited size. I'm fine with that as a constraint, but we should make it clear. Can you document the constraint on these functions?

Sources/_CryptoExtras/RSA/RSA_boring.swift
```diff
@@ -405,6 +440,33 @@ extension BoringSSLRSAPrivateKey {
             return _RSA.Signing.RSASignature(signatureBytes: output)
         }
 
+        fileprivate func decrypt<D: DataProtocol>(_ data: D, padding: _RSA.Encryption.Padding) throws -> _RSA.Encryption.RSADecryptedData {
+            let outputSize = Int(CCryptoBoringSSL_RSA_size(self.pointer))
+
+            let output = try Array<UInt8>(unsafeUninitializedCapacity: outputSize) { bufferPtr, length in
```
Lukasa, line 446: Same note here on the array.

Sources/_CryptoExtras/RSA/RSA.swift
gwynne: Hm... in that case, how would you feel about moving both key types into the `_RSA` "namespace" so that they're specific to neither subset (given that they are in fact nonspecific in terms of both implementation and usage).

Sources/_CryptoExtras/RSA/RSA.swift
gwynne: I was mimicking the design of `RSASignature` - if memory serves, I literally copy-pasted and changed the name.

Sources/_CryptoExtras/RSA/RSA_boring.swift
gwynne: This is another case where I was copy-pasting (from `signature()`, see below) and adapting to the purpose; doesn't it end up having the same problem, just slightly less explicitly?

Sources/_CryptoExtras/RSA/RSA.swift
Lukasa: It's still preferable to domain-separate the keys. While the key itself is not specific, using a key simultaneously for different purposes is almost always a red flag in a cryptosystem.

Sources/_CryptoExtras/RSA/RSA.swift
Lukasa: Yes, but RSA signatures are nominally at least somewhat structured. Encrypted bytes aren't.

Sources/_CryptoExtras/RSA/RSA_boring.swift
Lukasa: The existence of the RSASignature type is why the other side does this differently: given that we should just return a `Data` here, we can safely avoid the trouble.

Sources/_CryptoExtras/RSA/RSA.swift
gwynne: Ah, right... pardon my mental dust; it's been awhile since I dug into crypto 😅

Sources/_CryptoExtras/RSA/RSA_boring.swift
gwynne: My kingdom for `Data.init(unsafeUninitializedCapacity:initializingWith:)` 😂

gwynne:
@Lukasa I think I addressed everything you mentioned. I did have to add a `@_documentation(visibility: public)` attribute to the `_RSA` type to make the docs visible.

Sources/_CryptoExtras/RSA/RSA.swift
```diff
@@ -318,6 +324,133 @@ extension _RSA.Signing {
...
+    /// ## Common values:
+    ///
+    /// Key size|Padding|Max length
+    /// -|-|-
+    /// 2048|PKCS-OAEP|214 bytes
+    /// 3072|PKCS-OAEP|342 bytes
+    /// 4096|PKCS-OAEP|470 bytes
+    public func maximumEncryptSize(with padding: _RSA.Encryption.Padding) -> Int {
+        switch padding.backing {
+            case .pkcs1_oaep:
```
Lukasa, line 438: Nit: `case` indentation

Lukasa (commented):
Cool, this is very close to good.

Sources/_CryptoExtras/RSA/RSA.swift
gwynne: Fixed, with great disgust (see commit message) 😛

Lukasa (approved):
Nice, LGTM.

PR #128, opened 2022-08-22
Make BoringSSL wrapper match CryptoKit behaviour when working with x9.63

wrinkl3:
…63 representation

Make BoringSSL wrapper match CryptoKit behaviour when working with x9.63

### Checklist

### Motivation:

Recent versions of CryptoKit have changed how key initialization for EC x9.63 representation works. `P256.KeyAgreement.PublicKey(compressedRepresentation: Bytes)` now only accepts uncompressed form, while `P256.KeyAgreement.PublicKey(x963Representation: Bytes)` only accepts the compressed form. We'd like to bring that behavior in line with our BoringSSL wrappers.

### Modifications:

* Modified the `x963Representation`/`compressedRepresentation` wrappers to only accept compressed/uncompressed key length size
* Modified the unit tests to call `compressedRepresentation` when passing in compressed representation keys

### Result:

The BoringSSL wrapper behaviour now matches CryptoKit, and the unit tests now work with CryptoKit's latest version.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.4
+// swift-tools-version:5.7
 //===----------------------------------------------------------------------===//
 //
 // This source file is part of the SwiftCrypto open source project
@@ -56,7 +56,7 @@ if development {
 let package = Package(
     name: "swift-crypto",
     platforms: [
-        .macOS(.v10_15),
+        .macOS(.v13),
         .iOS(.v13),
         .watchOS(.v6),
         .tvOS(.v13),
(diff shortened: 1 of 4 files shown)


FredericJacobs:
Seems reasonable to me. @Lukasa ?

Lukasa (changes requested):
Thanks for this @wrinkl3, this looks great! Can I ask you to add some checks to `testCompressedKeys` to confirm that the `PublicKey.init(x963Representation:)` initialisers fail with compressed keys? I'd also love another test that is essentially the "other way", showing that `PublicKey.x963Representation` cannot be passed to `PublicKey.init(compressedRepresentation:)`.

Otherwise this looks great!

Lukasa:

Lukasa (approved):
Nice one, thanks!

PR #131, opened 2022-09-28
Raise minimum supported Swift version from 5.4 to 5.5

glbrntt:
Motivation:

Old Swift versions are periodically dropped. Now that 5.7 has been released, 5.4 will be dropped.

Modifications:

- Update tools version
- Remove 5.4 docker-compose
- Update 5.7 docker-compose to use released image and move from focal to jammy
- Update docs

Result:

Swift 5.5 is the minimum supported Swift version


docker/docker-compose.2204.57.yaml
```diff
@@ -0,0 +1,20 @@
+version: "3"
+
+services:
+
+  runtime-setup:
+    image: swift-crypto:22.04-5.7
+    build:
+      args:
+        base_image: "swift:5.7-jammy"
```
Lukasa, line 9: Please use `ubuntu_version`/`swift_version` in preference to `base_image`.

Lukasa (changes requested):

Lukasa (approved):

PR #133, opened 2022-10-13
Provide AES-GCM-SIV in CryptoExtras

FranzBusch:
# Motivation
BoringSSL exposes AES-GCM-SIV algorithms which are in general useful to have and provide a nonce-misuse resistant mode of AES-GCM. Since, `CryptoKit` is not exposing AES-GCM-SIV we need to add this to `_CryptoExtras`

# Modification
Exposes `AES-GCM-SIV` through `_CryptoExtras`.

# Result
We can now use `AES-GCM-SIV` through `_CryptoExtras`.


Lukasa:

FranzBusch:

Sources/CryptoBoringWrapper/AEAD/BoringSSLAEAD.swift
```diff
@@ -188,6 +190,63 @@ extension BoringSSLAEAD.AEADContext {
         let output = Data(bytesNoCopy: outputBuffer.baseAddress!, count: outputBuffer.count, deallocator: .free)
         return output
     }
+
+    /// An additional entry point for opening data where the ciphertext and the tag can be provided as one combined data . Covers the full gamut of types, including discontiguous data types. This must be inlinable.
+    @inlinable
+    public func open<Nonce: ContiguousBytes, AuthenticatedData: DataProtocol>(combinedCiphertextAndTag: Data, nonce: Nonce, authenticatedData: AuthenticatedData) throws -> Data {
```
FranzBusch, line 196: @Lukasa I have created a second interface here on the low level bits that takes the combined cipher and tag as the underlying BoringSSL API is expecting. The other approach would have been to only support one API and then we concatenate the cipher and tag for the SIV algorithms. I went with this approach since it avoids splitting the data and then putting it back together. Let me know what you think!

PR #137, opened 2022-12-24
build: adjust linking for Crypto

compnerd:
The package DLL `Crypto.dll` was building the BoringSSL wrapper (`CryptoBoringWrapper`) in the style that the user requested (static or shared).  However, there is a single user of the library and is not directly included by the SPM client.  Furthermore, the library was not setup for installation which makes redistribution of it impossible. Always build the library in a static mode to allow compaction into the `Crypto` target.

### DIFF
--- a/Sources/CryptoBoringWrapper/CMakeLists.txt
+++ b/Sources/CryptoBoringWrapper/CMakeLists.txt
@@ -12,7 +12,7 @@
 ##
 ##===----------------------------------------------------------------------===##
 
-add_library(CryptoBoringWrapper
+add_library(CryptoBoringWrapper STATIC
   "AEAD/BoringSSLAEAD.swift"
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @yim-lee @tomerd

yim-lee:
@compnerd Is this change required for https://github.com/apple/swift-package-manager/pull/5920?

compnerd:
I think that it will be needed, otherwise we will not be able to use the swift-package bits due to missing files.

yim-lee (approved):

yim-lee:
@Lukasa Can we merge this and cut a new release so we can update SwiftPM? Thanks.

FredericJacobs (approved):

Lukasa (approved):

Lukasa:
2.2.3 released.

Issue #139, opened 2023-01-07
Possible Over-Iteration in PrettyBytes


VaslD:
I was looking for a not `%02f` way of converting binary to hex `String`, and I stumbled upon the implementation in `PrettyBytes.swift`.

https://github.com/apple/swift-crypto/blob/9cc89f0170308b813af05dadcd26f9a2dee47713/Sources/Crypto/Util/PrettyBytes.swift#L45-L51

I'm not 100% sure about the `Sequence` conformance of non-contiguous `DataProtocol`, but it seemed to me that the `self` on line 46 should be the ignored parameter `$0` (or maybe eliminate the region logic since looping over `self` most likely would have taken non-contiguous memory regions into account?), otherwise the code reads:

> For **each segment** of the data buffer, loop over **the entirety of data buffer** and convert each byte to two hex chars.

Unless non-contiguous `DataProtocol` has non-stable indices and iterators, this code will over/re-iterate the data buffer.

Lukasa:
Yup, it sure does! Would you like to submit a patch to fix it? As a practical matter I think we can ignore the regions here, and just use the sequence conformance directly.

VaslD:
PR done. Don't know how to construct a multi-region `Data` to test it yet, but it shouldn't get worse. :-)

PR #141, opened 2023-01-10
Fix over-iterating buffer in hex string utility.

VaslD:
Fixes https://github.com/apple/swift-crypto/issues/139: over-iteration in hex string.

### Checklist

### Motivation:

Upon discussing it in https://github.com/apple/swift-crypto/issues/139, an issue was identified that could cause unexpected hex output for a data buffer whose memory spans multiple non-contiguous regions.

### Modifications:

Removed memory region specific logic. Fix applied to 3 copies of the same function.

Also added unit test on Apple platforms to compare with `NSString`'s hex output.

### Result:

With multi-region `DataProtocol`s, `hexString` should return correct hex.

### DIFF
--- a/Tests/CryptoTests/Utils/PrettyBytesTests.swift
+++ b/Tests/CryptoTests/Utils/PrettyBytesTests.swift
@@ -1,0 +1,33 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftCrypto open source project
+//
+// Copyright (c) 2019-2020 Apple Inc. and the SwiftCrypto project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.md for the list of SwiftCrypto project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+import Foundation
+import XCTest
+
+// Not sure if NSString-bridged "String.init(format:_:)" is available on Linux/Windows.
+#if (os(macOS) || os(iOS) || os(tvOS) || os(watchOS)) && CRYPTO_IN_SWIFTPM_FORCE_BUILD_API
+
+@testable import Crypto
+
+class PrettyBytesTests: XCTestCase {
+    func testHexString() {
+        let random = Data((1...64).map { _ in UInt8.random(in: UInt8.min...UInt8.max)})
+
+        let hexString = random.hexString
+        let target = random.map { String(format: "%02x", $0) }.joined()
+
+        XCTAssertEqual(hexString, target)
+    }
+}
+
+#endif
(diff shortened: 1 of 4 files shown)


Tests/CryptoTests/Utils/PrettyBytesTests.swift
```diff
@@ -0,0 +1,33 @@
...
+// See CONTRIBUTORS.md for the list of SwiftCrypto project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+import Foundation
+import XCTest
+
+// Not sure if NSString-bridged "String.init(format:_:)" is available on Linux/Windows.
+#if (os(macOS) || os(iOS) || os(tvOS) || os(watchOS)) && CRYPTO_IN_SWIFTPM_FORCE_BUILD_API
```
Lukasa, line 18: Sorry, this line needs to be:

```
#if (os(macOS) || os(iOS) || os(watchOS) || os(tvOS)) && CRYPTO_IN_SWIFTPM && !CRYPTO_IN_SWIFTPM_FORCE_BUILD_API
// Skip tests that require @testable imports of CryptoKit.
#else
#if (os(macOS) || os(iOS) || os(watchOS) || os(tvOS)) && !CRYPTO_IN_SWIFTPM_FORCE_BUILD_API
@testable import CryptoKit
#else
@testable import Crypto
#endif
```

Lukasa (changes requested):
Thanks for this! One minor tweak in the compiler defines and we should be go.

Tests/CryptoTests/Utils/PrettyBytesTests.swift
VaslD: I did look at what other tests wrote. But I wasn't checking whether to use `CryptoKit` or `Crypto`, rather I needed to know (guess) whether `String.init(format:_:)` was available in the SDK. On appleOS this API is provided for sure by bridging from `NSString`'s initializer, but on Linux and Windows I couldn't know.

The code you suggested would make this test always available cross-platform. Let me verify if Swift's cross-platform `Foundation` port has this API. Otherwise I'll need to do something like `sprintf` to make it cross-platform.

VaslD: Also `CryptoTests` defines its own copy of `PrettyBytes`. It's weird because I though `@testable import` means that test target has access to everything non-private in the original `Crypto` module.

This would mean the `PrettyBytesTests` does not actually need to import anything, as the test only tests functions defined in the test target itself.

Can you clarify this?

Lukasa: I'm happy with only testing the one in the tests module, tbh.

PR #142, opened 2023-01-16
Add AES block function to CryptoExtras

Lukasa:
In some cases developers have need of AES modes other than what we support today, or need the ability to implement algorithms that require access to the core AES permutation function. As a starting point, we can provide access to the AES block function in CryptoExtras.

### DIFF
--- a/Sources/_CryptoExtras/AES/Block Function.swift
+++ b/Sources/_CryptoExtras/AES/Block Function.swift
@@ -21,10 +21,28 @@ import Foundation
 extension AES {
     private static let blockSize = 128 / 8
 
+    /// Apply the AES permutation operation in the encryption direction.
+    ///
+    /// This function applies the core AES block operation to `payload` in the encryption direction. Note that this is
+    /// not performing any kind of block cipher mode, and does not authenticate the payload. This is a dangerous primitive
+    /// that should only be used to compose higher-level primitives, and should not be used directly.
+    ///
+    /// - parameter payload: The payload to encrypt. Must be exactly 16 bytes long.
+    /// - parameter key: The encryption key to use.
+    /// - throws: On invalid parameter sizes.
     public static func _permute<Payload: MutableCollection>(_ payload: inout Payload, key: SymmetricKey) throws where Payload.Element == UInt8 {
         return try Self.permuteBlock(&payload, key: key, permutation: .forward)
     }
 
+    /// Apply the AES permutation operation in the decryption direction.
+    ///
+    /// This function applies the core AES block operation to `payload` in the decryption direction. Note that this is
+    /// not performing any kind of block cipher mode, and does not authenticate the payload. This is a dangerous primitive
+    /// that should only be used to compose higher-level primitives, and should not be used directly.
+    ///
+    /// - parameter payload: The payload to decrypt. Must be exactly 16 bytes long.
+    /// - parameter key: The decryption key to use.
+    /// - throws: On invalid parameter sizes.
     public static func _inversePermute<Payload: MutableCollection>(_ payload: inout Payload, key: SymmetricKey) throws where Payload.Element == UInt8 {
         return try Self.permuteBlock(&payload, key: key, permutation: .backward)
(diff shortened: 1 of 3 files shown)


Sources/_CryptoExtras/AES/Block Function.swift
```diff
@@ -0,0 +1,125 @@
...
+
+extension AES {
+    private static let blockSize = 128 / 8
+
+    public static func _permute<Payload: MutableCollection>(_ payload: inout Payload, key: SymmetricKey) throws where Payload.Element == UInt8 {
+        return try Self.permuteBlock(&payload, key: key, permutation: .forward)
+    }
+
+    public static func _inversePermute<Payload: MutableCollection>(_ payload: inout Payload, key: SymmetricKey) throws where Payload.Element == UInt8 {
+        return try Self.permuteBlock(&payload, key: key, permutation: .backward)
```
FranzBusch, line 29: Even though these are "private". Should we add some docs?

FranzBusch (approved):
Looks good. Just one comment w.r.t. docs

Sources/_CryptoExtras/AES/Block Function.swift
Lukasa: Sure, one sec.

Lukasa: Added, please re-review.

FranzBusch (approved):

Sources/_CryptoExtras/AES/Block Function.swift
```diff
@@ -0,0 +1,143 @@
...
+    /// Apply the AES permutation operation in the encryption direction.
+    ///
+    /// This function applies the core AES block operation to `payload` in the encryption direction. Note that this is
+    /// not performing any kind of block cipher mode, and does not authenticate the payload. This is a dangerous primitive
+    /// that should only be used to compose higher-level primitives, and should not be used directly.
+    ///
+    /// - parameter payload: The payload to encrypt. Must be exactly 16 bytes long.
+    /// - parameter key: The encryption key to use.
+    /// - throws: On invalid parameter sizes.
+    public static func permute<Payload: MutableCollection>(_ payload: inout Payload, key: SymmetricKey) throws where Payload.Element == UInt8 {
```
FredericJacobs, line 33: What makes `MutableCollection` the right protocol here?

Sources/_CryptoExtras/AES/Block Function.swift
```diff
@@ -0,0 +1,143 @@
...
+    /// Apply the AES permutation operation in the decryption direction.
+    ///
+    /// This function applies the core AES block operation to `payload` in the decryption direction. Note that this is
+    /// not performing any kind of block cipher mode, and does not authenticate the payload. This is a dangerous primitive
+    /// that should only be used to compose higher-level primitives, and should not be used directly.
+    ///
+    /// - parameter payload: The payload to decrypt. Must be exactly 16 bytes long.
+    /// - parameter key: The decryption key to use.
+    /// - throws: On invalid parameter sizes.
+    public static func inversePermute<Payload: MutableCollection>(_ payload: inout Payload, key: SymmetricKey) throws where Payload.Element == UInt8 {
```
FredericJacobs, line 46: I'm not sure yet they need to be separate functions as opposed to a bool to define the direction.

Sources/_CryptoExtras/AES/Block Function.swift
Lukasa: Assuming we want to support in-place modification, `MutableCollection` is the simplest protocol that allows us to do what we want. Specifically, `MutableCollection` adds the `set` accessor to the subscript, so we can modify the elements in the `Collection`. It also has a fast-path available to it for collections that can expose their storage.

Almost all other protocols in this space that you can think of refine the `MutableCollection` protocol in some way: for example, `DataProtocol` refines `MutableCollection`.

FredericJacobs (approved):

PR #145, opened 2023-02-08
Support SwiftWasm

AndrewBarba:
Adds support for WASI platform

### Checklist

### Motivation:

WASM is growing in popularity and we now have the ability to compile Swift application to WASM via the SwiftWasm toolchain. We need a performant Crypto implementation that compiles to WASM.

### Modifications:

I've modified `Package.swift` to define certain flags when building for `Platform.wasi`. I've also explicitly excluded 3 files from BoringSSL which rely on `netdb.h` which we cannot use in WASM. It is safe to exclude these files for all platforms as nothing in Crypto relies on the networking features of BoringSSL.

### Result:

This project is now able to be included in SwiftWasm applications.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -47,6 +47,7 @@ if development {
         Platform.linux,
         Platform.android,
         Platform.windows,
+        Platform.wasi,
     ]
     dependencies = [
         .target(name: "CCryptoBoringSSL", condition: .when(platforms: platforms)),
@@ -86,6 +87,12 @@ let package = Package(
                  * only.  Unconditionally define it instead.
                  */
                 .define("WIN32_LEAN_AND_MEAN"),
+                /*
+                 * These defines are required on WASM, to disable use of pthread.
+                 */
+                .define("ARCH_WASM32", .when(platforms: [Platform.wasi])),
+                .define("OPENSSL_NO_THREADS_CORRUPT_MEMORY_AND_LEAK_SECRETS_IF_THREADED", .when(platforms: [Platform.wasi])),
+                .define("OPENSSL_NO_ASM", .when(platforms: [Platform.wasi])),
             ]
         ),
         .target(
--- a/Package.swift
+++ b/Package.swift
@@ -95,7 +95,7 @@ let package = Package(
                  */
                 .define("WIN32_LEAN_AND_MEAN"),
                 /*
-                 * These defines are required on WASM, to disable use of pthread.
+                 * These defines are required on Wasm/WASI, to disable use of pthread.
                  */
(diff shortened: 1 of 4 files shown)


jpope19 (approved):

Sajjon:
Fantastic! 🥳😍

AndrewBarba:
Here's a GitHub Actions test run on the SwiftWasm 5.7 toolchain: https://github.com/swift-cloud/swift-crypto/actions/runs/4130755431/jobs/7137745847

Package.swift
```diff
@@ -86,6 +94,11 @@ let package = Package(
                  * only.  Unconditionally define it instead.
                  */
                 .define("WIN32_LEAN_AND_MEAN"),
+                /*
+                 * These defines are required on WASM, to disable use of pthread.
```
Lukasa, line 98: How will this play with web assembly threads? Do we have some way to stay up to speed with whether those are supported in Swift Wasm/WASI? cc @MaxDesiatov

Lukasa (commented):
I'm tentatively happy with this proposal modulo my question about threading. I'd also like to page @MaxDesiatov to give his feedback.

Lukasa:

Package.swift
MaxDesiatov: WASI-libc has some rudimentary support for pthread API, but we shouldn't assume it's available, I'm not sure SwiftWasm has that enabled. Even when it is, with very high probability it may just use noop stubs under the hood. Very likely we may need a separate triple altogether for multi-threaded Wasm.

Package.swift
```diff
@@ -78,6 +79,13 @@ let package = Package(
                 "hash.txt",
                 "include/boringssl_prefix_symbols_nasm.inc",
                 "CMakeLists.txt",
+                /*
+                 * These files are excluded to support WASM which cannot include <netdb.h>.
```
MaxDesiatov, line 83: nit: it's WASI that provides these headers, not Wasm.

```suggestion
                 * These files are excluded to support WASI libc which doesn't provide <netdb.h>.
```

Package.swift
MaxDesiatov, line 98: wording nit, also following capitalization defined in the WebAssembly spec
```suggestion
                 * These defines are required on Wasm/WASI, to disable use of pthread.
```

MaxDesiatov (approved):
LGTM, thanks!

Lukasa:

Lukasa (approved):
Nice one, this patch LGTM!

PR #152, opened 2023-03-14
Exclude AES from CMake

yim-lee:
SwiftPM doesn't need it and this would allow removal of `CCryptoBoringSSL` dependency on non-Darwin platforms, thus reducing SwiftPM size.

~~Also remove `CryptoKitErrors_boring.swift` in `_CryptoExtras` because the same util already exists in `Crypto` and code in `_CryptoExtras` can use that instead.~~


yim-lee:
Need to add exclusions to `update_cmakelists` or else those files would keep getting re-added to `CMakeLists`

Lukasa (approved):

Issue #168, opened 2023-04-10
QUIC Crypto Primitives Support


btoms20:
#### Request:

 - Add support for AES-ECB under the `Insecure` umbrella struct 
   - Used to generate the mask required for AES based Header Protection [RFC 9001 - Section 5.4.3](https://datatracker.ietf.org/doc/html/rfc9001#name-aes-based-header-protection)
 - Add support for setting the 4 byte `Counter` alongside the `Nonce`/`IV` when initializing a `ChaCha20` `Cipher`.
   - Used to generate the mask required for ChaCha20 based Header Protection [RFC 9001 - Section 5.4.4](https://datatracker.ietf.org/doc/html/rfc9001#name-chacha20-based-header-prote)
 
 #### Importance:

 Adds the missing cryptographic functions required to support the QUIC protocol as defined in [RFC 9001](https://datatracker.ietf.org/doc/html/rfc9001)

FredericJacobs:
Hey @btoms20,

Thanks for this proposal. As @Lukasa said, we swift-crypto already provides the AES permutation function, from which you can do one-block ECB since you only need a single block for header protection.

Regarding the proposal to expose counters, the challenge is that `counter` is really an internal construct of `ChaCha20` (which we don't officially support with POLY1305 authentication).

The `swift-crypto` API is really designed to be misuse resistant and exposing the counter through an API is clearly very risky, so we would likely make this some sort of an `Insecure`-prefixed extension function.

However, as it currently stands, we are not supporting stream ciphers without authentication. And therefore, it would be a bit of a stretch to have a way to set the counter on top of our existing `AEAD` API.

But we can keep this request to revisit this if we have any developments on  unauthenticated stream ciphers.

btoms20:
@Lukasa, thanks for pointing that out!

@FredericJacobs, sounds good and understood. Hopefully it can make its way into the Insecure API in the future.

Thanks for quick responses!

btoms20:
I had completely overlooked _CryptoExtras. Maybe if / when the Insecure ChaCha20 Counter extension gets implemented, it could be tucked away in _CryptoExtras with a not-so-public API and all of the appropriate / necessary warnings. That might help deter your average user from unknowingly misusing it.

Lukasa:
[...]

PR #169, opened 2023-04-11
ChaCha20 CTR Encryption

btoms20:
Added a method to interact with BoringSSL's `CCryptoBoringSSL_CRYPTO_chacha_20()`

### Checklist

### Motivation:

Having the ability to explicitly set the Counter in the ChaCha20 Cipher allows us to generate 'Header Protection Masks' as described in [RFC 9001 - Using TLS to Secure QUIC](https://datatracker.ietf.org/doc/html/rfc9001#name-chacha20-based-header-prote) and issue #168 

### Modifications:

Added a single encryption method under the `Insecure.ChaCha20CTR` enum that lets the user of this library perform single block encryptions with an explicitly defined Counter and Nonce / IV.

The ChaCha20CTR enum described above is defined in the `_CryptoExtras` product.

### Result:

After including both `Crypto` and `_CryptoExtras` in your project you'll have access to the ChaCha20CTR enum and the associated encryption method that enables one off encryption operations.
```swift
import Crypto
import _CryptoExtras

Insecure.ChaCha20CTR.encrypt(
   message: DataProtocol,
   using: SymmetricKey,
   counter: Insecure.ChaCha20CTR.Counter,
   nonce: Insecure.ChaCha20CTR.Nonce
) throws -> Data { ... }
```
See the included [ChaCha20CTRTests.swift](Tests/_CryptoExtrasTests/ChaCha20CTRTests.swift) file for usage examples.


Sources/_CryptoExtras/ChaCha20CTR/BoringSSL/ChaCha20CTR_boring.swift
```diff
@@ -0,0 +1,54 @@
...
+        }
+        guard nonce.count == Insecure.ChaCha20CTR.nonceByteCount else {
+            throw CryptoKitError.incorrectParameterSize
+        }
+
+        let plaintext = Array(message)
+        var ciphertext = Array<UInt8>(repeating: 0, count: plaintext.count)
+        let nonce = Array<UInt8>(nonce)
+
+        self.chacha20CTR(out: &ciphertext, plaintext: plaintext, inLen: plaintext.count, key: key.withUnsafeBytes { Array($0) }, nonce: nonce, counter: counter)
```
Lukasa, line 34: Rather than create a temporary Array, can you wrap this call in `key.withUnsafeBytes` and pass the closure argument as the `key` parameter?

Sources/_CryptoExtras/ChaCha20CTR/BoringSSL/ChaCha20CTR_boring.swift
```diff
@@ -0,0 +1,54 @@
...
+
+        return ciphertext
+    }
+
+    static func chacha20CTR(out: UnsafeMutablePointer<UInt8>, plaintext: UnsafePointer<UInt8>, inLen: Int, key: UnsafePointer<UInt8>, nonce: UnsafePointer<UInt8>, counter: UInt32) {
+        let outPtr = UnsafeMutableRawPointer(out).assumingMemoryBound(to: UInt8.self)
+        let inPtr = UnsafeRawPointer(plaintext).assumingMemoryBound(to: UInt8.self)
+        let keyPtr = UnsafeRawPointer(key).assumingMemoryBound(to: UInt8.self)
+        let noncePtr = UnsafeRawPointer(nonce).assumingMemoryBound(to: UInt8.self)
```
Lukasa, line 43: Why are we erasing and rebinding the pointers here?

Sources/_CryptoExtras/ChaCha20CTR/BoringSSL/ChaCha20CTR_boring.swift
```diff
@@ -0,0 +1,54 @@
...
+    static func encrypt<M: DataProtocol, N: DataProtocol>(key: SymmetricKey, message: M, counter: UInt32, nonce: N) throws -> [UInt8] {
+        guard key.bitCount == Insecure.ChaCha20CTR.keyBitsCount else {
+            throw CryptoKitError.incorrectKeySize
+        }
+        guard nonce.count == Insecure.ChaCha20CTR.nonceByteCount else {
+            throw CryptoKitError.incorrectParameterSize
+        }
+
+        let plaintext = Array(message)
+        var ciphertext = Array<UInt8>(repeating: 0, count: plaintext.count)
```
Lukasa, line 31: Rather than return `Array`, can we return `Data`? This matches the existing API surface.

Sources/_CryptoExtras/ChaCha20CTR/BoringSSL/ChaCha20CTR_boring.swift
```diff
@@ -0,0 +1,54 @@
...
+        let nonce = Array<UInt8>(nonce)
+
+        self.chacha20CTR(out: &ciphertext, plaintext: plaintext, inLen: plaintext.count, key: key.withUnsafeBytes { Array($0) }, nonce: nonce, counter: counter)
+
+        return ciphertext
+    }
+
+    static func chacha20CTR(out: UnsafeMutablePointer<UInt8>, plaintext: UnsafePointer<UInt8>, inLen: Int, key: UnsafePointer<UInt8>, nonce: UnsafePointer<UInt8>, counter: UInt32) {
```
Lukasa, line 39: Can we take `plaintext` as an `UnsafeBufferPointer` instead of as a pointer directly?

Sources/_CryptoExtras/ChaCha20CTR/BoringSSL/ChaCha20CTR_boring.swift
```diff
@@ -0,0 +1,54 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftCrypto open source project
+//
+// Copyright (c) 2022 Apple Inc. and the SwiftCrypto project authors
```
Lukasa, line 5:
```suggestion
// Copyright (c) 2023 Apple Inc. and the SwiftCrypto project authors
```

Sources/_CryptoExtras/ChaCha20CTR/ChaCha20CTR.swift
```diff
@@ -0,0 +1,45 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftCrypto open source project
+//
+// Copyright (c) 2022 Apple Inc. and the SwiftCrypto project authors
```
Lukasa, line 5:
```suggestion
// Copyright (c) 2023 Apple Inc. and the SwiftCrypto project authors
```

Sources/_CryptoExtras/ChaCha20CTR/ChaCha20CTR.swift
```diff
@@ -0,0 +1,45 @@
...
+        /// - Parameters:
+        ///   - message: The message to encrypt
+        ///   - key: A 256-bit encryption key
+        ///   - counter: A 4 byte counter (UInt32), defaults to 0
+        ///   - nonce: A 12 byte nonce for ChaCha20 encryption. The nonce must be unique for every use of the key to seal data.
+        /// - Returns: The encrypted ciphertext
+        /// - Throws: CipherError errors
+        /// - Warning: You most likely want to use the ChaChaPoly implemention with AuthenticatedData available at `Crypto.ChaChaPoly`
+        public static func encrypt<Plaintext: DataProtocol, Nonce: DataProtocol>
+        (_ message: Plaintext, using key: SymmetricKey, counter: UInt32 = 0, nonce: Nonce) throws -> [UInt8] {
```
Lukasa, line 41: Can we use a typed `Nonce` here, as with the other constructions, rather than have it as a general bag of bytes?

Lukasa (changes requested):
Thanks for this! I've left some notes in the diff.

Sources/_CryptoExtras/ChaCha20CTR/ChaCha20CTR.swift
btoms20: I went ahead and created both a ChaCha20CTR.Nonce and ChaCha20CTR.Counter struct https://github.com/apple/swift-crypto/pull/169/commits/56aff30f6cdafd051bd32b7e1cb99d006c16cf54

Sources/_CryptoExtras/ChaCha20CTR/BoringSSL/ChaCha20CTR_boring.swift
btoms20: Sorry about that, pointers confuse me sometimes... Fixed in https://github.com/apple/swift-crypto/pull/169/commits/1c7dcfc7b603f7b4ea4274ac2a6b5ad99c7f962e

Sources/_CryptoExtras/ChaCha20CTR/BoringSSL/ChaCha20CTR_boring.swift
btoms20: Hopefully https://github.com/apple/swift-crypto/pull/169/commits/1c7dcfc7b603f7b4ea4274ac2a6b5ad99c7f962e is what you had in mind? I tried to reference other calls into BoringSSL throughout the code to keep the style consistent.

Sources/_CryptoExtras/ChaCha20CTR/BoringSSL/ChaCha20CTR_boring.swift
```diff
@@ -0,0 +1,56 @@
...
+@_implementationOnly import CryptoBoringWrapper
+import Foundation
+
+enum OpenSSLChaCha20CTRImpl {
+    static func encrypt<M: DataProtocol, N: ContiguousBytes>(key: SymmetricKey, message: M, counter: UInt32, nonce: N) throws -> Data {
+        guard key.bitCount == Insecure.ChaCha20CTR.keyBitsCount else {
+            throw CryptoKitError.incorrectKeySize
+        }
+
+        var ciphertext = Array<UInt8>(repeating: 0, count: message.count)
```
Lukasa, line 27: We can use `Data` directly here rather than copying to a new `Data` at the end. You'll have to replace the `&ciphertext` with a call to `withUnsafeMutableBytes`, but that's ok, and it avoids an extra allocation and copy.

Sources/_CryptoExtras/ChaCha20CTR/BoringSSL/ChaCha20CTR_boring.swift
```diff
@@ -0,0 +1,56 @@
...
+
+        key.withUnsafeBytes { keyPtr in
+            nonce.withUnsafeBytes { noncePtr in
+                message.withContiguousStorageIfAvailable { plaintext in
```
Lukasa, line 31: This call isn't quite right. We need to do a bit of a dance here. I recommend looking at the way the `seal` function works, where there is an outer layer that optionally flattens the `DataProtocol` before calling an inner layer that is generic over `ContiguousBytes`. That's what we want to do here.

btoms20: Okay hopefully https://github.com/apple/swift-crypto/pull/169/commits/17d166b425bbdf5b0a4d9fba30a55ee19bb58635 implements this correctly?

Sources/_CryptoExtras/ChaCha20CTR/ChaCha20CTR.swift
```diff
@@ -0,0 +1,110 @@
...
+        public init<D: DataProtocol>(data: D) throws {
+            if data.count != Insecure.ChaCha20CTR.counterByteCount {
+                throw CryptoKitError.incorrectParameterSize
+            }
+
+            if data.regions.count == 1 {
+                self.counter = data.regions.first!.withUnsafeBytes { $0.load(as: UInt32.self) }
+            } else {
+                self.counter = Array(data).withUnsafeBytes { $0.load(as: UInt32.self) }
+            }
```
Lukasa, line 98: Ok, we can avoid some unsafe code here.

All `DataProtocol` are also `Collection<UInt8>`, so we can safely load this using a safe method. Additionally, it'll solve the endianness problem that this code has, where it claims to use little-endian but actually uses native-endian.

Let's replace it with:

```swift
if data.count != Insecure.ChaCha20CTR.counterByteCount {
    throw CryptoKitError.incorrectParameterSize
}

let startIndex = data.startIndex
self.counter = (
    (UInt32(data[data.index(startIndex, offsetBy: 0)]) << 0) |
    (UInt32(data[data.index(startIndex, offsetBy: 1)]) << 8) |
    (UInt32(data[data.index(startIndex, offsetBy: 2)]) << 16) |
    (UInt32(data[data.index(startIndex, offsetBy: 3)]) << 24) |
)
```

Tests/_CryptoExtrasTests/ChaCha20CTRTests.swift
```diff
@@ -0,0 +1,99 @@
...
+import _CryptoExtras
+
+class ChaCha20CTRTests: XCTestCase {
+
+    /// Test Vector - https://datatracker.ietf.org/doc/html/rfc9001#name-chacha20-poly1305-short-hea
+    func testChaCha20CTR_v1() throws {
+        let hpKey = try Array(hexString: "25a282b9e82f06f21f488917a4fc8f1b73573685608597d0efcb076b0ab7a7a4")
+        /// Sample = 0x5e5cd55c41f69080575d7999c25a5bfb
+        let counterAsData = try Array(hexString: "5e5cd55c")
+        let counterAsUInt32 = counterAsData.withUnsafeBytes { $0.load(as: UInt32.self) }
```
Lukasa, line 26: Rather than use a load, can we write the integer hex literal here? That avoids having symmetric bugs in the loading and product code.

Tests/_CryptoExtrasTests/ChaCha20CTRTests.swift
```diff
@@ -0,0 +1,99 @@
...
+        XCTAssertEqual(mask, try Data(hexString: "aefefe7d03"))
+        XCTAssertEqual(mask, mask2)
+    }
+
+    /// Test Vector - https://www.ietf.org/archive/id/draft-ietf-quic-v2-10.html#name-chacha20-poly1305-short-head
+    func testChaCha20CTR_v2() throws {
+        let hpKey = try Array(hexString: "d659760d2ba434a226fd37b35c69e2da8211d10c4f12538787d65645d5d1b8e2")
+        /// Sample = 0xe7b6b932bc27d786f4bc2bb20f2162ba
+        let counterAsData = try Array(hexString: "e7b6b932")
+        let counterAsUInt32 = counterAsData.withUnsafeBytes { $0.load(as: UInt32.self) }
```
Lukasa, line 41: Same note here, let's use an integer literal.

Lukasa (commented):
Great, this is getting really close to ready to land. I've left a few more notes, but they're getting very trivial now.

Sources/_CryptoExtras/ChaCha20CTR/ChaCha20CTR.swift
btoms20: Nice! 👍

Tests/_CryptoExtrasTests/ChaCha20CTRTests.swift
btoms20: In order to keep the hex orientation I needed to use UInt32(bigEndian:) initializer. Is that alright? Or would you like me to convert the hex into it's actual decimal value and use that instead?

btoms20:
```swift
let counterAsUInt32 = UInt32(bigEndian: 0x5e5cd55c)
// or
let counterAsUInt32 = UInt32(1557486686)
```

Lukasa:

Lukasa:

Lukasa:
Excellent, one quick fixup: can you run `scripts/update_cmakelists.sh` and commit the changes?

Lukasa:

btoms20:
Awesome! Thanks for guiding me through this PR, I really appreciate all the comments / help. Sorry for taking up so much of your time! 😅

PR #172, opened 2023-04-13
Drop support for Swift 5.5

Lukasa:
This patch drops support for Swift 5.5.


yim-lee:
> Can we remove the 5.5 builder here please?

@Lukasa Done.

yim-lee (approved):

PR #177, opened 2023-04-18
Set CRYPTO_IN_SWIFTPM

etcwilde:
Needed to expose SecureEncalve through SwiftCrypto when building for Swift-Certificates. CRYPTO_IN_SWIFTPM is set all of the time in package.swift, while CRYPTO_IN_SWIFTPM_FORCE_BUILD_API is only set in development. Lets enable the one that's on all of the time, and remove the one that isn't so that we get our declarations.

Since `CRYPTO_IN_SWIFTPM` is enabled all of the time, would it make sense to not have it? Or what exactly is it used for?

### DIFF
--- a/Sources/Crypto/CMakeLists.txt
+++ b/Sources/Crypto/CMakeLists.txt
@@ -82,7 +82,7 @@ add_library(Crypto
   "Util/Zeroization.swift")
 
 target_compile_definitions(Crypto PRIVATE
-  CRYPTO_IN_SWIFTPM_FORCE_BUILD_API)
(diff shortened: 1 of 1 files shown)


yim-lee:

Lukasa (approved):
Thanks @etcwilde. Yeah, we should set `CRYPTO_IN_SWIFTPM` unconditionally. We use it as part of our source management strategy to allow the API files of CryptoKit to be shared between this implementation and CryptoKit. It's one of the signals we use to decide whether we should define an API here or defer to CryptoKit.

PR #180, opened 2023-04-22
Add inequality operator ( != ) for digest and DataProtocol comparison

joey-gm:
Add an inequality comparison operator to determine whether a digest is not equal to a collection of contiguous bytes.

### Checklist

### Motivation:

The equal ( == ) and not equal ( != ) operators are available for comparing two digests. However, only the equal operator is available for the comparison between a digest and a collection of contiguous bytes.

This modification adds the missing inequality operator ( != ) for digest and DataProtocol comparison.

### Modifications:

The following static method has been added to **_Digest.swift_** as an extension to **Digest**.

`public static func != <D: DataProtocol>(lhs: Self, rhs: D) -> Bool`

A corresponding test case has also been added to **_DigestsTests.swift_**.

### Result:

Instead of checking `if !(digest == someDataBytes) {...}`, we can now write `if digest != someDataBytes {...}`.

### DIFF
--- a/Sources/Crypto/Digests/Digest.swift
+++ b/Sources/Crypto/Digests/Digest.swift
@@ -11,10 +11,11 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 //===----------------------------------------------------------------------===//
+import Foundation
+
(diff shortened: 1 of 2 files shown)


Lukasa (changes requested):
Thanks for this! As with several other PRs open on this repo, any change to the public API surface requires synchronization with CryptoKit in Apple's SDKs. While this is a reasonable change, I'm going to defer merging it for now.

PR #194, opened 2023-08-06
Fix up flags used for Docker

gwynne:
The `-t`, `-I`, and `--privileged` flags for `docker run` are not necessary here. Also adds logic to use the same version of Swift as the macOS installation (note: breaks use with beta versions of Xcode, as the Docker tag for such Swift versions won't exist yet).

### DIFF
--- a/scripts/vendor-boringssl.sh
+++ b/scripts/vendor-boringssl.sh
@@ -101,7 +101,7 @@ function mangle_symbols {
         )
 
         # Now cross compile for our targets.
-        docker run --rm -v$(pwd):/src -w/src --platform linux/arm64 swift5.8-jammy \
+        docker run --rm -v$(pwd):/src -w/src --platform linux/arm64 swift:5.8-jammy \
             swift build --product CCryptoBoringSSL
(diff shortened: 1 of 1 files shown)


Lukasa (changes requested):
Can you drop the changes to use the same version of Swift as macOS? The Swift version isn't really important here so we can use whatever we like.

gwynne:
> Can you drop the changes to use the same version of Swift as macOS? The Swift version isn't really important here so we can use whatever we like.

Done!

scripts/vendor-boringssl.sh
```diff
@@ -101,9 +101,9 @@ function mangle_symbols {
         )
 
         # Now cross compile for our targets.
-        docker run -t -i --rm --privileged -v$(pwd):/src -w/src --platform linux/arm64 swift:5.8-jammy \
+        docker run --rm -v$(pwd):/src -w/src --platform linux/arm64 swift5.8-jammy \
```
Lukasa, line 104: Missing colon.

gwynne: Fixed, sorry 😅

Lukasa (approved):

PR #198, opened 2023-08-17
Use `SecKeyGetBlockSize` instead of `kSecAttrKeySizeInBits`

dnadoba:
### Checklist

### Motivation:
Security.framework has change to return the exact bit count of the key size instead of the block size in bytes in macOS 14 beta and aligned releases. However, BoringSSL and the previous implementation relyed on that.
`SecKeyCopyAttributes ` is also rather expensive to call and `keySizeInBits` is accessed immediately (twice) after initialisation. 

### Modifications:

call `SecKeyGetBlockSize(_:)`` instead of looking up the `kSecAttrKeySizeInBits` in the attributes dictionary.

### Result:

test pass again and allocations are reduced.


Lukasa (approved):

Lukasa:

PR #203, opened 2023-10-10
pkcs8PEMRepresenation for RSA private keys

bjhomer:
This PR adds a `pkcs8PEMRepresentation` property to RSA public keys.

### Checklist

- [ ] ~~I've run `.script/generate_boilerplate_files_with_gyb` and included updated generated files in a commit of this pull request~~

### Motivation:

There are two widely-used formats for public and private keys. [PKCS#1](https://www.rfc-editor.org/rfc/rfc3447#appendix-A.1) defines the format for RSA keys, which begin with `BEGIN RSA {PUBLIC/PRIVATE KEY}` 

Later, as additional key types were implemented, more general containers were introduced. [X.509](https://www.rfc-editor.org/rfc/rfc3280#section-4.1) introduces `SubjectPublicKeyInfo` (SPKI), which defines a general structure for containing _any_ type of public key (RSA or otherwise). [PKCS#8](https://www.rfc-editor.org/rfc/rfc5208#page-3) does the same for private keys. `privateKey.pemRepresentation` gives the PKCS#1 representation, but there's no equivalent for getting the PCKS8 representation.

Our implementation of RSA keys included a way to get both the new and old format of public keys (via `publicKey.pemRepresentation` and `publicKey.pkcs1PEMRepresentation`), but we didn't have a way to get both the new and old formats of _private_ keys.

This PR introduces `_RSA.PrivateKey.pkcs8PEMRepresentation` to fill that gap.

### Modifications:

- `_RSA`'s backing private key types have gained a new `pkcs8PEMRepresentation` property.
- I adjusted the package file to pass the `CRYPTO_IN_SWIFTPM` and `CRYPTO_IN_SWIFTPM_FORCE_BUILD_API` flags to the `_CryptoExtras` target as well. 
- I updated the implementation of `_CryptoExtras` to check for those flags, rather than checking for `#if canImport(Security)`. This made it easier to test the RSA changes when building on macOS.

Note that the BoringSSL implementation of `pkcs8PEMRepresentation` includes a trailing newline. The naive implementation using Security.framework would not have that trailing newline, but for consistency between the two implementations, I have added it to both.

I'm open to instead stripping off the trailing new line on the BoringSSL implementation, but I was less sure of how to implement that.

### Result:

RSA private keys now have `pkcs8PEMRepresentation`.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -130,7 +130,8 @@ let package = Package(
             ],
             exclude: [
                 "CMakeLists.txt",
-            ]
+            ],
+            swiftSettings: swiftSettings
         ),
         .target(
             name: "CryptoBoringWrapper",
--- a/Sources/_CryptoExtras/RSA/RSA.swift
+++ b/Sources/_CryptoExtras/RSA/RSA.swift
@@ -144,6 +144,10 @@ extension _RSA.Signing {
             self.backing.pemRepresentation
         }
 
+        public var pkcs8PEMRepresentation: String {
+            self.backing.pkcs8PEMRepresentation
+        }
+
         public var keySizeInBits: Int {
             self.backing.keySizeInBits
         }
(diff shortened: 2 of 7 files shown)


Lukasa:

Lukasa (approved):
Nice change, I really like this. Thanks!

PR #205, opened 2023-10-19
BoringSSLRSAPublicKey use EVP_PKEY API

rnro:
Switch BoringSSLRSAPublicKey encrypt/decrypt methods to use `EVP_PKEY_*` API

### Motivation

Increased versatility of the `BoringSSLRSAPublicKey` encrypt and decrypt methods to make future extensions possible.

### Modifications

Switch the `BoringSSLRSAPublicKey` encrypt and decrypt methods to use the `EVP_PKEY_*` API directly rather than going through the RSA abstractions.

### Result

Increased versatility of the `BoringSSLRSAPublicKey` encrypt and decrypt methods.

### DIFF
--- a/Sources/_CryptoExtras/RSA/RSA_boring.swift
+++ b/Sources/_CryptoExtras/RSA/RSA_boring.swift
@@ -122,17 +122,20 @@ extension BoringSSLRSAPublicKey {
         }
 
         fileprivate init(copying other: Backing) {
-            self.pointer = CCryptoBoringSSL_RSAPublicKey_dup(other.pointer)
+            self.pointer = CCryptoBoringSSL_EVP_PKEY_new()
+            let rsaPublicKey = CCryptoBoringSSL_RSAPublicKey_dup(CCryptoBoringSSL_EVP_PKEY_get0_RSA(other.pointer))
+            CCryptoBoringSSL_EVP_PKEY_assign_RSA(self.pointer, rsaPublicKey)
         }
 
         fileprivate init(pemRepresentation: String) throws {
             var pemRepresentation = pemRepresentation
+            self.pointer = CCryptoBoringSSL_EVP_PKEY_new()
 
             // There are two encodings for RSA public keys: PKCS#1 and the SPKI form.
             // The SPKI form is what we support for EC keys, so we try that first, then we
             // fall back to the PKCS#1 form if that parse fails.
             do {
-                self.pointer = try pemRepresentation.withUTF8 { utf8Ptr in
+                let rsaPublicKey = try pemRepresentation.withUTF8 { utf8Ptr in
                     return try BIOHelper.withReadOnlyMemoryBIO(wrapping: utf8Ptr) { bio in
                         guard let key = CCryptoBoringSSL_PEM_read_bio_RSA_PUBKEY(bio, nil, nil, nil) else {
                             throw CryptoKitError.internalBoringSSLError()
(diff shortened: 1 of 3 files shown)


Sources/_CryptoExtras/RSA/RSA_boring.swift
```diff
@@ -259,13 +259,32 @@ extension BoringSSLRSAPublicKey {
...
+                    // nil engine defaults to the standard implementation with no hooks
+                    let ctx = CCryptoBoringSSL_EVP_PKEY_CTX_new(pkey, nil)
+                    defer {
+                        CCryptoBoringSSL_EVP_PKEY_CTX_free(ctx)
+                    }
+
+                    CCryptoBoringSSL_EVP_PKEY_encrypt_init(ctx)
+                    CCryptoBoringSSL_EVP_PKEY_CTX_set_rsa_padding(ctx, rawPadding)
+
+                    var writtenLength = bufferPtr.count
```
rnro, line 279: We don't currently do anything with this - we could use it to check if the length meets our expectations and precondition/throw. WDYT @Lukasa ?

Sources/_CryptoExtras/RSA/RSA_boring.swift
```diff
@@ -465,26 +484,46 @@ extension BoringSSLRSAPrivateKey {
...
                         bufferPtr.baseAddress,
-                        self.pointer,
-                        rawPadding
+                        &writtenLength,
+                        dataPtr.baseAddress,
+                        dataPtr.count
                     )
-                    return rc
+
+                    return rc == 0 ? CInt(-1) : CInt(writtenLength)
```
Lukasa, line 520: Let's not use -1 as a sentinel, let's just throw.

Sources/_CryptoExtras/RSA/RSA_boring.swift
rnro: https://commondatastorage.googleapis.com/chromium-boringssl-docs/evp.h.html#EVP_PKEY_encrypt_init
> EVP_PKEY_encrypt encrypts in_len bytes from in. If out is NULL, the maximum size of the ciphertext is written to out_len. Otherwise, *out_len must contain the number of bytes of space available at out. If sufficient, the ciphertext will be written to out and *out_len updated with the true length.

My interpretation of the above was that `out_len`/`writtenLength` should contain the buffer length and not 0.

Lukasa: Nice catch.

Sources/_CryptoExtras/RSA/RSA_boring.swift
```diff
@@ -259,13 +259,32 @@ extension BoringSSLRSAPublicKey {
                     switch padding.backing {
                     case .pkcs1_oaep: rawPadding = RSA_PKCS1_OAEP_PADDING
                     }
-                    let rc = CCryptoBoringSSLShims_RSA_public_encrypt(
-                        CInt(dataPtr.count),
-                        dataPtr.baseAddress,
+
+                    let pkey = CCryptoBoringSSL_EVP_PKEY_new()
```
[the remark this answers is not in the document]
rnro: Obviously you have more context here but I'm not sure that's the best thing to do. 

The RSA pointer is currently used in a few different locations outside of the initializers e.g. `pkcs1DERRepresentation`, `derRepresentation`, `keySizeInBits`, `isValidSignature`. Each of these would then need a call to get the RSA pointer unless we also stored that (which is an option). I looked for alternative functions but there are a few including `BIO` operations which now have annotations in the docs which look like deprecations ([here](https://commondatastorage.googleapis.com/chromium-boringssl-docs/rsa.h.html#i2d_RSAPublicKey)) which I think make this complex enough to warrant a separate piece of work if we want to do it.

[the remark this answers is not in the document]
rnro: Since the `get0` functions offer a cheap way to access the underlying RSA object we have opted to just keep hold of the EVP_KEY

Lukasa (approved):
Nice one, this LGTM.

PR #206, opened 2023-10-19
Add support for pkcs1 oaep sha256

rnro:
Add support for pkcs1 oaep sha256

### Motivation

Add RSA OAEP-sha256 support in addition to the increasingly aged SHA-1
implementation.

### Modifications

* Add a `Digest` enum to `_RSA.Encryption.Padding`
* Add a digest associated type to the `.pkcs1_oaep` padding enum case
  to allow us to distinguish different digest hash functions.
* Add a `PKCS1_OAEP_SHA256` public static let to allow users to use the
  new hash function.
* Enable SHA-256 RSA encryption tests

### Result

* Support for RSA OAEP-sha256

**NOTE:** This change sticks with the BoringSSL default behaviour and the case which uses the new SHA-256 digest hash function also uses SHA-256 as its mask generation function.

### DIFF
--- a/Sources/_CryptoExtras/RSA/RSA.swift
+++ b/Sources/_CryptoExtras/RSA/RSA.swift
@@ -401,11 +401,11 @@ extension _RSA.Encryption {
 extension _RSA.Encryption {
     public struct Padding {
         internal enum Backing {
-            case pkcs1_oaep
+            case pkcs1_oaep(Digest)
         }
-        
+
         internal var backing: Backing
-        
+
         private init(_ backing: Backing) {
             self.backing = backing
         }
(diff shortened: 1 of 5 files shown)


Sources/_CryptoExtras/RSA/RSA.swift
```diff
@@ -401,19 +401,32 @@ extension _RSA.Encryption {
...
+    internal enum Digest {
+        case md4
+        case md5
+        case sha1
+        case sha224
+        case sha256
+        case sha384
+        case sha512
+        case sha512_256
+        case blake2b256
```
Lukasa, line 429: This is excessive, let's just have only the digests we actually support.

Tests/_CryptoExtrasTests/TestRSAEncryption.swift
```diff
@@ -36,8 +36,13 @@ final class TestRSAEncryption: XCTestCase {
 
         let derPubKey = derPrivKey.publicKey
 
-        guard group.sha == "SHA-1", group.mgfSha == "SHA-1" else {
-            // We currently only support SHA-1 OAEP, which is very legacy but oh well.
+        let padding: _RSA.Encryption.Padding
```
Lukasa, line 39: We should add a new set of test vectors: https://github.com/google/wycheproof/blob/master/testvectors/rsa_oaep_2048_sha256_mgf1sha256_test.json

Lukasa (changes requested):
I've left some notes, please also update RSA_security.swift.

Lukasa (approved):

Lukasa (changes requested):
Sorry, hang-on: we need a Security.framework equivalent.

dnadoba (approved):

Lukasa (approved):
Nice one @rnro, thanks!

Issue #209, opened 2023-11-14
How to set CBC padding behavior?


lovetodream:
### Question Checklist

- [x] I'm using the latest version of Swift Crypto (master branch)
- [x] I read the [Contribution Guidelines](https://github.com/apple/swift-crypto/blob/master/CONTRIBUTING.md)
- [x] I searched for [existing GitHub issues](https://github.com/apple/swift-crypto/issues)

### Question Subject
I'm currently using CryptoSwift for CBC de-/encryption and I'd like to move that to swift-crypto.
For my use-case I have to disable automatic padding, is that possible?

#### Question Description
Currently I have the following code to decrypt a CBC encrypted payload sent from an Oracle db server during authenticating:
```swift
func decryptCBC(_ key: [UInt8], _ encryptedText: [UInt8]) throws -> [UInt8] {
    let iv = [UInt8](repeating: 0, count: 16)
    let aes = try AES(key: key, blockMode: CBC(iv: iv), padding: .noPadding)
    return try aes.decrypt(encryptedText)
}
```

I tried migrating it to swift-crypto and now it does automatic padding and fails with `CryptoKitError.incorrectParameterSize`, when trying to trim the padding:
```swift
func decryptCBC(_ key: [UInt8], _ encryptedText: [UInt8]) throws -> Data {
    let iv = [UInt8](repeating: 0, count: 16)
    return try AES._CBC.decrypt(encryptedText, using: .init(data: key), iv: AES._CBC.IV(ivBytes: iv)) // noPadding equivalent?
}
```

I guess I'll have the same problem on encryption.
```swift
// current (CryptoSwift):
func encryptCBC(_ key: [UInt8], _ plainText: [UInt8], zeros: Bool = false) throws -> [UInt8] {
    var plainText = plainText
    let blockSize = 16
    let iv = [UInt8](repeating: 0, count: blockSize)
    let n = blockSize - plainText.count % blockSize
    if n != 0, !zeros {
        plainText += Array<UInt8>(repeating: UInt8(n), count: n)
    }
    let aes = try AES(key: key, blockMode: CBC(iv: iv), padding: zeros ? .zeroPadding : .noPadding)
    var encryptor = try aes.makeEncryptor()
    return try encryptor.update(withBytes: plainText, isLast: true) + encryptor.finish()
}

// new (swift-crypto):
func encryptCBC(_ key: [UInt8], _ plainText: [UInt8], zeros: Bool = false) throws -> Data {
    var plainText = plainText
[...]

PR #210, opened 2023-11-16
Add support for CBC without padding

lovetodream:
This PR adds support for CBC de- and encryption without padding, as discussed in #209 

### Checklist

### Motivation:

As described in #209, I personally need this to migrate an oracle driver from a third party crypto lib to swift-crypto. I think other users might benefit from this addition too.

### Modifications:

I've added an overload to the `encrypt` and `decrypt` methods of `AES._CBC`, allowing the user to configure if padding should be added or not. With `noPadding` set to `true`, an error will be thrown if the `plaintext` isn't a multiple of the block size. I've added the corresponding inline documentation.

I've also added tests to ensure both encrypting and decrypting without padding work as expected. Although those tests might not be sufficient enough, because I couldn't find good resources online. I've created a bunch of random hex strings and encrypted/decrypted them using another implementation of paddingless CBC and checked if I receive the expected results. To further validate the feature, I've tested it as part of the authentication in oracle-nio, which worked in all test scenarios I've been running.

### Result:

After merging this, it will be possible to use CBC without padding.

### DIFF
--- a/Sources/_CryptoExtras/AES/AES_CBC.swift
+++ b/Sources/_CryptoExtras/AES/AES_CBC.swift
@@ -40,11 +40,29 @@ extension AES {
         ///   - iv: The initialization vector.
         /// - Returns: The encrypted ciphertext.
         public static func encrypt<Plaintext: DataProtocol>(_ plaintext: Plaintext, using key: SymmetricKey, iv: AES._CBC.IV) throws -> Data {
+            try self.encrypt(plaintext, using: key, iv: iv, noPadding: false)
+        }
+        
+        /// Encrypts data using AES-CBC.
+        ///
+        /// - Parameters:
+        ///   - plaintext: The message to encrypt.
+        ///   - key: A encryption key.
+        ///   - iv: The initialization vector.
+        ///   - noPadding: If set to `true`, padding won't be added.
+        /// - Returns: The encrypted ciphertext.
+        ///
+        /// - Note: If `noPadding` is set to `true`, `plainText` has to be a multiple of the blockSize (16 bytes). Otherwise an error will be thrown.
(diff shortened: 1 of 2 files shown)


Lukasa:

Sources/_CryptoExtras/AES/AES_CBC.swift
```diff
@@ -84,6 +102,18 @@ extension AES {
...
+        
+        /// Decrypts data using AES-CBC.
+        ///
+        /// - Parameters:
+        ///   - ciphertext: The encrypted ciphertext.
+        ///   - key: A decryption key.
+        ///   - iv: The initialization vector.
+        ///   - noPadding: If this is set to `true`, padding won't be removed.
+        /// - Returns: The decrypted message.
+        public static func decrypt<Ciphertext: DataProtocol>(_ ciphertext: Ciphertext, using key: SymmetricKey, iv: AES._CBC.IV, noPadding: Bool = false) throws -> Data {
```
Lukasa, line 116: Let's not default the value here.

Lukasa (changes requested):
Generally looks good, got one small note for you.

Sources/_CryptoExtras/AES/AES_CBC.swift
lovetodream: Yeah, I don't know how that made it into the commit :), changed it now

Lukasa:

lovetodream:
> One final suggestion here: these tests encrypt and decrypt only inputs that are a single block in size, or that are not an even multiple. Can you add a few tests that use multi-block-size inputs?

Sure, I've added a few more with 2-16x block sizes

lovetodream:
Anything left for me to do @Lukasa?

Lukasa:
Hi @lovetodream, sorry about the delay, I've been swamped. I'll take a look now.

Lukasa:

Lukasa (approved):

Lukasa:
Nice change, thanks for your hard work!

Issue #208, opened 2023-11-06
Add Sendable annotation to Crypto Extras


0xTim:
#### Motivation:

Use Swift Crypto types in concurrency contexts invariably starts to hit issues as no types are Sendable. I'm aware there are issues around the main Crypto target as that has match CryptoKit (or does it? Could we add Sendable conformance on Linux?) but it would be great to add it to the types in Crypto Extras

#### Importance:

It's possible to achieve this with a mix of `@preconcurrency` and `@unchecked Sendable` but neither of these are ideal for obvious reasons.

Lukasa:
Yes, the goal is that the Crypto API exactly matches the CryptoKit one. This avoids users needing to deal with platform-specific imports or other logic.

For CryptoExtras, this can definitely be done separately. @rnro want to take a look at CryptoExtras?

rnro:
Yep, sure.

0xTim:
Fixed in #211

PR #211, opened 2023-12-11
Add Sendable annotations to Crypto Extras

rnro:
### Motivation

Add Sendable annotations to Crypto Extras to reflect current sandability so that downstream packages can implement sendable conformance.

### Modifications

* Add Sendable conformance to appropriate public symbols.
* RSA public and private keys are notable omissions, this is because they are backed by SecKey from security framework and we have no guarantees about their sendability

### Result

* Accurate Sendable annotations


dnadoba:

dnadoba (approved):

rnro:
See https://github.com/apple/swift-crypto/issues/208

PR #213, opened 2023-12-19
Update availability annotations in tests

simonjbeaumont:
### Checklist

#### ~If you've made changes to `gyb` files~
- [ ] ~I've run `.script/generate_boilerplate_files_with_gyb` and included updated generated files in a commit of this pull request~

### Motivation:

When building the tests for iOS and tvOS, the build fails because we are using APIs that aren't available on all platform versions. E.g.:

```
Tests/CryptoTests/Digests/DigestsTests.swift:117:37: error: 'blockByteCount' is only available in iOS 13.2 or newer
        XCTAssertEqual(Insecure.MD5.blockByteCount, 64)
```

### Modifications:

- Add `@available` guard to test that uses `blockByteCount` API.
- Update `DataProtocol.hexString` to not use `String.init(unsafeUninitializedCapacity:initializingUTF8With:)`, which is only available on iOS 14+, and instead use `Array.init(unsafeUninitializedCapacity:initializingWith:)`, which is available on older platforms.
 
### Result:

Tests can now build for and run on iOS and tvOS.

### DIFF
--- a/Tests/_CryptoExtrasTests/Utils/BytesUtil.swift
+++ b/Tests/_CryptoExtrasTests/Utils/BytesUtil.swift
@@ -60,7 +60,7 @@ extension DataProtocol {
     var hexString: String {
         get {
             let hexLen = self.count * 2
-            return String.init(unsafeUninitializedCapacity: hexLen) { buf in
+            let bytes = Array(unsafeUninitializedCapacity: hexLen) { buf, count in
                 var offset = 0
 
                 self.regions.forEach { (_) in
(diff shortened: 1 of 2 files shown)


Tests/_CryptoExtrasTests/Utils/BytesUtil.swift
```diff
@@ -70,8 +70,9 @@ extension DataProtocol {
                         offset += 1
                     }
                 }
-                return offset
+                count = offset
             }
+            return String(bytes: bytes, encoding: .ascii)!
```
Lukasa, line 75: No need to use this Foundation API here, we can instead use `String(decoding: bytes, as: UTF8.self)`.

simonjbeaumont: Thanks! Updated.

Lukasa (approved):

Lukasa:

PR #214, opened 2023-12-19
Add more Sendable annotations to Crypto Extras

rnro:
### Motivation

Add Sendable annotations to Crypto Extras to reflect current sandability so that downstream packages can implement sendable conformance.

### Modifications

* Add Sendable conformance to some missed public symbols.
* RSA public and private keys are now marked sendable with their backing storage on darwin OS marked unchecked sendable until SecKey from security framework has sendability annotations

### Result

* More accurate Sendable annotations

### DIFF
--- a/Sources/_CryptoExtras/AES/AES_CBC.swift
+++ b/Sources/_CryptoExtras/AES/AES_CBC.swift
@@ -18,7 +18,7 @@ import Foundation
 extension AES {
     /// The Advanced Encryption Standard (AES) Cipher Block Chaining (CBC) cipher
     /// suite.
-    public enum _CBC {
+    public enum _CBC: Sendable {
         private static var blockSize: Int { 16 }
 
         private static func encryptBlockInPlace(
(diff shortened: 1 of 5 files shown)


Sources/_CryptoExtras/AES/AES_CBC.swift
```diff
@@ -18,7 +18,7 @@ import Foundation
 extension AES {
     /// The Advanced Encryption Standard (AES) Cipher Block Chaining (CBC) cipher
     /// suite.
-    public enum _CBC {
+    public enum _CBC: Sendable {
```
Lukasa, line 21: This enum is uninhabited, it doesn't need a `Sendable` annotation.

Sources/_CryptoExtras/AES/AES_GCM_SIV.swift
```diff
@@ -25,7 +25,7 @@ import Foundation
 /// Types associated with the AES GCM SIV algorithm
 extension AES.GCM {
     /// AES in GCM SIV mode with 128-bit tags.
-    public enum _SIV {
+    public enum _SIV: Sendable {
```
Lukasa, line 28: Same here.

Lukasa (changes requested):

Sources/_CryptoExtras/AES/AES_CBC.swift
rnro: Ah, yep, good catch, I missed that on this pass.

Lukasa (approved):

PR #218, opened 2024-02-14
Update lower bound for RSA key sizes

ptoffy:
Update lower bound for RSA key sizes to 2048 rather than 1024

### Checklist
- [x] I've updated the documentation if necessary (didn't find any documentation to update about this)

### Motivation:

Since 2015, NIST recommends a minimum key size of 2048 bits as stated in https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-57pt1r5.pdf
This is also stated in the JWA specification (RFC7518 [3.3](https://datatracker.ietf.org/doc/html/rfc7518#section-3.3) and [3.5](https://datatracker.ietf.org/doc/html/rfc7518#section-3.5)) and in the [RSA SSH spec](https://datatracker.ietf.org/doc/html/rfc8332#section-5.1), that's why this PR enforces the use of key sizes `>=2048` rather than `>=1024` as before.

### Modifications:

Update lower bound for RSA key sizes to 2048 rather than 1024, specifically in the RSA key initialisers. Also update tests which use 1024 bits-sized keys and make sure they throw

### Result:

Creating RSA keys of sizes `<2048` will throw


Lukasa (changes requested):
Thanks for this change! However, as written I don’t think we should take it. 

RSA support is not provided for new uses, but for compatibility with old ones. To that compatibility end, some support for weaker key sizes is likely to be necessary.  

However, I do think we could make the case for this change _if_ we introduced an explicitly marked `unsafe` or `unchecked` initializer that reinstated the current bounds. Are you open to that?

ptoffy:
@Lukasa that seems sensible. I've updated the PR with those change, let me know what you think!

Lukasa (approved):

PR #232, opened 2024-05-21
extras: Add APIs for RSA Blind Signatures

simonjbeaumont:
### Motivation

RFC 9474 defines the RSA Blind Signatures protocol[^0], which are a useful building block of privacy-preserving schemes.

### Modifications

This PR adds the following public API surface to implement the RSA Blind Signatures protocol:

```swift
public struct _RSA.BlindSigning.Parameters<H: HashFunction> {}

extension _RSA.BlindSigning.Parameters where H == SHA384 {
    public static let RSABSSA_SHA384_PSS_Randomized: Self<SHA384>
    public static let RSABSSA_SHA384_PSS_Deterministic: Self<SHA384>
    public static let RSABSSA_SHA384_PSSZERO_Randomized: Self<SHA384>
    public static let RSABSSA_SHA384_PSSZERO_Deterministic: Self<SHA384>
}

public struct _RSA.BlindSigning.PrivateKey {
    public typealias Parameters = _RSA.BlindSigning.Parameters<H>

    // ... similar initializers as other protocol-segreated key types, but taking parameters parameter ...

    // ... same properties as other protocol-segregated key types ...

    public func blindSignature<D: DataProtocol>(for message: D) throws -> _RSA.BlindSigning.BlindSignature
}

public struct _RSA.BlindSigning.PublicKey {
    public typealias Parameters = _RSA.BlindSigning.Parameters<H>

    // ... similar initializers as other protocol-segreated key types, but taking parameters parameter ...

    // ... same properties as other protocol-segregated key types ...

    public func prepare<D: DataProtocol>(_ message: D) -> _RSA.BlindSigning.PreparedMessage

    public func blind(_ message: _RSA.BlindSigning.PreparedMessage) throws -> (blindedMessage: Data, blindInverse: _RSA.BlindSigning.BlindInverse)

    public func finalize(_ signature: _RSA.BlindSigning.BlindSignature, for message: _RSA.BlindSigning.PreparedMessage,
 blindInverse: _RSA.BlindSigning.BlindInverse) throws -> _RSA.Signing.RSASignature

    public func isValidSignature(_ signature: _RSA.Signing.RSASignature, for message: _RSA.BlindSigning.PreparedMessage) -> Bool
}
```

It also adds tests using the test vectors from the RFC, where possible. That is, for each of the operations _except_ for `blind`, which we don't have the BoringSSL APIs that would allow us to inject the fixed salt value from the test vectors.

While it's possible to implement the server-side operations using Security framework, it does not expose the APIs we would need to implement the client-side operations so, because the goal is to also provide the client-side operations in a subsequent PR, the implementation uses BoringSSL on all platforms.

In order to construct the RSA keys from the test vector parameters, some BoringSSL helpers were added.

### Result

New API in the `_CryptoExtras` module for the  the RSA Blind Signatures protocol as defined in RFC 9474, with support for all named variants:  for the `RSABSSA_SHA384_PSS_Randomized`, `RSABSSA_SHA384_PSS_Deterministic`, `RSABSSA_SHA384_PSSZERO_Randomized`, and `RSABSSA_SHA384_PSSZERO_Deterministic`.

[^0]: https://datatracker.ietf.org/doc/rfc9474/


Sources/_CryptoExtras/RSA/RSA.swift
```diff
@@ -246,6 +246,7 @@ extension _RSA.Signing {
         internal enum Backing {
             case pkcs1v1_5
             case pss
+            case pssZero
```
FredericJacobs, line 249: Can we ensure that other callers would not use this by adding a comment on this enum case or having a more deterring name?

Sources/_CryptoExtras/RSA/RSA+BlindSigning.swift
```diff
@@ -0,0 +1,393 @@
...
+
+extension _RSA.BlindSigning.PrivateKey {
+    /// Generate a blind signature with the given key for a blinded message.
+    ///
+    /// - Parameter blindedMessage: The blinded message to sign.
+    /// - Returns: A blind signature.
+    /// - Throws: If there is a failure producing the signature.
+    ///
+    /// - Seealso: [RFC 9474: BlindSign](https://www.rfc-editor.org/rfc/rfc9474.html#name-blindsign).
+    /// - TODO: Should this accept a new `BlindedMesage` type to help guide protocol usage?
```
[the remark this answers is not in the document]
simonjbeaumont: I assume this applies to the similar comments on the other would-be APIs, e.g. `PreparedMessage`?

Sources/_CryptoExtras/Util/BoringSSLHelpers.swift
```diff
@@ -93,4 +92,158 @@ extension FixedWidthInteger {
...
+    }
+
+    func modulo(_ mod: Self, nonNegative: Bool = false) -> Self {
+        var result = BIGNUM()
+
+        let rc = withUnsafeMutablePointer(to: &result) { resultPtr in
+            withUnsafePointer(to: self) { selfPtr in
+                withUnsafePointer(to: mod) { modPtr in
+                    let bnCtx = CCryptoBoringSSL_BN_CTX_new()!
+                    defer { CCryptoBoringSSL_BN_CTX_free(bnCtx) }
```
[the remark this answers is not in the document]
simonjbeaumont: Can do. My plan is that this code is very short lived. Once we have SPI for ArbitraryPrecisionInteger, most of these functions will be implemented in terms of that type.

Sources/_CryptoExtras/RSA/RSA_boring.swift
```diff
@@ -101,6 +101,10 @@ extension BoringSSLRSAPrivateKey {
     internal func decrypt<D: DataProtocol>(_ data: D, padding: _RSA.Encryption.Padding) throws -> Data {
         return try self.backing.decrypt(data, padding: padding)
     }
+
+    internal func blindSignature<D: DataProtocol>(_ blindedMessage: D) throws -> _RSA.BlindSigning.BlindSignature {
```
FredericJacobs, line 105: If we're aligning with public API for signatures, we would use `blindSignature(for message)`

As far as the signer goes, they do a blind sign operation, but they don't know what they sign, so I would just call it the message, not blindedMessage

Tests/_CryptoExtrasTests/TestRSABlindSigning.swift
```diff
@@ -0,0 +1,99 @@
...
+        for testVector in filteredTestVectors {
+            // Prepare
+            do {
+                let message = try Data(hexString: testVector.msg)
+                switch testVector.parameters.preparation {
+                case .identity:
+                    let preparedMessage = _RSA.BlindSigning.prepare(message, parameters: testVector.parameters)
+                    XCTAssertEqual(Data(preparedMessage), message)
+                case .randomized:
+                    break  // Until we have SPI secure bytes.
```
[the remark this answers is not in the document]
simonjbeaumont: Until we have SPI for `SecureBytes` the randomised prepare function fatal errors as unimplemented. The prepare operation is not public yet so this is just a placeholder test.

Sources/_CryptoExtras/RSA/RSA_boring.swift
```diff
@@ -564,6 +591,73 @@ extension BoringSSLRSAPrivateKey {
...
+                        &outputCount,
+                        signatureBufferPtr.baseAddress,
+                        signatureBufferPtr.count,
+                        blindedMessageBufferPtr.baseAddress,
+                        blindedMessageBufferPtr.count,
+                        RSA_NO_PADDING
+                    ) == 1 else {
+                        throw CryptoKitError.internalBoringSSLError()
+                    }
+                    precondition(outputCount == signatureByteCount)
```
FredericJacobs, line 619: Did you test this on enough key sizes that it's always true? (eg. if first bytes are 0, they are not getting trimmed)

Just want to make sure we don't crash if we have an oddly formatted private key generated some day.

Tests/_CryptoExtrasTests/TestRSABlindSigning.swift
```diff
@@ -0,0 +1,99 @@
...
+
+    static func load(from fileURL: URL) throws -> [Self] {
+        let json = try Data(contentsOf: fileURL)
+        let decoder = JSONDecoder()
+        return try decoder.decode([Self].self, from: json)
+    }
+}
+
+final class TestRSABlindSigning: XCTestCase {
+    func testAgainstRFC9474TestVectors() throws {
```
FredericJacobs, line 37: Can we add test cases that ensure that we can go through the whole flow? 

In the case of test vectors, we skip the blinding step, so it would be good to verify that if we invoke all methods, we still have correctness. Would be also interesting to generate non-byte aligned key-sizes since for RSA those are possible.

FredericJacobs (approved):

Sources/_CryptoExtras/RSA/RSA.swift
simonjbeaumont: This is an internal backing enum and not exposed in the API. Are you suggesting we go further and annotate it in some fashion to deter accidental _internal_ usage?

Sources/_CryptoExtras/RSA/RSA_boring.swift
simonjbeaumont: Can I confirm if this comment pertains to only the parameter name or to the type. There’s some other discussion about whether this takes its own `BlindedMessage` type rather than arbitrary data.

Sources/_CryptoExtras/RSA/RSA_boring.swift
simonjbeaumont: Ah. Let me double check on this, thanks!

Tests/_CryptoExtrasTests/TestRSABlindSigning.swift
simonjbeaumont: We cannot in this PR in its current form since we do not have any of the client-side functions implemented. This PR adds only blindSign and verify.

Sources/_CryptoExtras/RSA/RSA.swift
FredericJacobs: Better safe than sorry, because deterministic signatures are not at all desirable in many other settings.

Sources/_CryptoExtras/RSA/RSA_boring.swift
simonjbeaumont: Done: the public API and all the internal functions are now `blindSignature(for message:)`.

Sources/_CryptoExtras/RSA/RSA+BlindSigning.swift
```diff
@@ -0,0 +1,416 @@
...
+        public init(pemRepresentation: String, parameters: Parameters = .RSABSSA_SHA384_PSS_Randomized) throws {
+            self.backing = try BackingPrivateKey(pemRepresentation: pemRepresentation)
+            self.parameters = parameters
+
+            guard self.keySizeInBits >= 2048 else {
+                throw CryptoKitError.incorrectParameterSize
+            }
+        }
+
+        /// Construct an RSA public key from a PEM representation.
```
Lukasa, line 133:
```suggestion
        /// Construct an RSA private key from a PEM representation.
```

Sources/_CryptoExtras/RSA/RSA+BlindSigning.swift
```diff
@@ -0,0 +1,416 @@
...
+        public init<Bytes: DataProtocol>(derRepresentation: Bytes, parameters: Parameters = .RSABSSA_SHA384_PSS_Randomized) throws {
+            self.backing = try BackingPrivateKey(derRepresentation: derRepresentation)
+            self.parameters = parameters
+
+            guard self.keySizeInBits >= 2048 else {
+                throw CryptoKitError.incorrectParameterSize
+            }
+        }
+
+        /// Construct an RSA public key from a DER representation.
```
Lukasa, line 160:
```suggestion
        /// Construct an RSA private key from a DER representation.
```

Lukasa (approved):

Tests/_CryptoExtrasTests/TestRSABlindSigning.swift
simonjbeaumont: Now we've included the client operations in this PR, we've been able to add full end-to-end flow tests for all named variants and supported key sizes.

simonjbeaumont:
@Lukasa this is ready for another pass when you have time.

Sources/_CryptoExtras/RSA/RSA+BlindSigning.swift
```diff
@@ -0,0 +1,414 @@
...
+    /// randomized message preparation.
+    ///
+    /// - Seealso: [RFC 9474: RSABSSA Variants](https://www.rfc-editor.org/rfc/rfc9474.html#name-rsabssa-variants).
+    public static let RSABSSA_SHA384_PSSZERO_Deterministic = Self<SHA384>(padding: .PSSZERO, preparation: .identity)
+}
+
+extension _RSA.BlindSigning {
+    /// An input ready to be blinded, possibly prepended with random bytes.
+    ///
+    /// Users should not attempt to create values of this type; it is created and returned by the prepare operation.
```
Lukasa, line 309: There's no should not here, they cannot.

Sources/_CryptoExtras/RSA/RSA+BlindSigning.swift
```diff
@@ -0,0 +1,414 @@
...
+    public struct PreparedMessage {
+        var rawRepresentation: Data
+    }
+
+    /// The blinding inverse for a blinded messaeg, used to unblind a blind signature.
+    ///
+    /// Users should not attempt to create values of this type; it is created and returned by the blind operation.
```
Lukasa, line 316: There's no should not here, they cannot.

Sources/_CryptoExtras/RSA/RSA+BlindSigning.swift
```diff
@@ -0,0 +1,414 @@
...
+        try self.backing.blindSignature(for: message)
+    }
+}
+
+extension _RSA.BlindSigning.PublicKey {
+    /// Prepare a message to be signed using the blind signing protocol.
+    ///
+    /// - Parameter message: The message to be signed.
+    /// - Parameter parameters: Parameters used in the blind signing protocol.
+    /// - Returns: A preprared mesage, modified according to the parameters provided.
```
Lukasa, line 340:
```suggestion
    /// - Returns: A prepared message, modified according to the parameters provided.
```

Sources/_CryptoExtras/RSA/RSA+BlindSigning.swift
```diff
@@ -0,0 +1,414 @@
...
+    ///
+    /// - Seealso: [RFC 9474: Prepare](https://www.rfc-editor.org/rfc/rfc9474.html#name-prepare).
+    public func prepare<D: DataProtocol>(_ message: D) -> _RSA.BlindSigning.PreparedMessage {
+        switch self.parameters.preparation {
+        case .identity:
+            return _RSA.BlindSigning.PreparedMessage(rawRepresentation: Data(message))
+        case .randomized:
+            return _RSA.BlindSigning.PreparedMessage(rawRepresentation: Data(SystemRandomNumberGenerator.randomBytes(count: 32)) + message)
```
Lukasa, line 348: This is a fairly inefficient construction. We're going to do the following:

1. Allocate a 32-byte array.
2. Attempt to append a message to it, which will require allocating a new Array.
3. Allocate a Data and copy that array into it.

A better algorithm is to allocate a Data, reserve 32 bytes + message.count, and then append the random bytes and the message. That will remove two allocations.

While we're here, can we add `@inlinable` on our generic functions?

Sources/_CryptoExtras/RSA/RSA+BlindSigning.swift
```diff
@@ -0,0 +1,414 @@
...
+
+    /// Blind a message to be signed by the server using the blind signing protocol.
+    ///
+    /// - Parameter message: The message to be signed.
+    /// - Returns: The blinded message, and its inverse for unblinding its blind signature.
+    ///
+    /// - Seealso: [RFC 9474: Blind](https://www.rfc-editor.org/rfc/rfc9474.html#name-blind).
+    public func blind(
+        _ message: _RSA.BlindSigning.PreparedMessage
+    ) throws -> (blindedMessage: Data, blindInverse: _RSA.BlindSigning.BlindInverse) {
```
Lukasa, line 360: I'm generally opposed to using tuples as return values, and would prefer we used a nominal type here.

Sources/_CryptoExtras/RSA/RSA+BlindSigning.swift
```diff
@@ -0,0 +1,414 @@
...
+            return self.backing.isValidSignature(signature, for: H.hash(data: message.rawRepresentation), padding: .PSSZERO)
+        }
+    }
+}
+
+extension _RSA.BlindSigning {
+    /// Errors defined in the RSA Blind Signatures protocol.
+    ///
+    /// - Seealso: [RFC 9474: Errors](https://www.rfc-editor.org/rfc/rfc9474.html#name-errors).
+    enum Error: Swift.Error {
```
Lukasa, line 404: This error should probably be public, and probably not be an enum.

Sources/_CryptoExtras/RSA/RSA_boring.swift
```diff
@@ -315,6 +344,215 @@ extension BoringSSLRSAPublicKey {
...
+
+            // 1. encoded_msg = EMSA-PSS-ENCODE(msg, bit_len(n)) with Hash, MGF, and salt_len as defined in the parameters
+            // 2. If EMSA-PSS-ENCODE raises an error, re-raise the error and stop
+            let hashDigestType = try! DigestType(forDigestType: H.Digest.self)
+            let hash = H.hash(data: message.rawRepresentation)
+            let saltLength: Int32
+            switch parameters.padding {
+            case .PSS: saltLength = Int32(H.Digest.byteCount)
+            case .PSSZERO: saltLength = 0
+            }
```
Lukasa, line 384: It might be nice to factor this out to a method on the parameters.

Sources/_CryptoExtras/RSA/RSA_boring.swift
```diff
@@ -315,6 +344,215 @@ extension BoringSSLRSAPublicKey {
...
+            var encodedMessage = [UInt8](repeating: 0, count: outputSize)
```
Lukasa, line 385: We could use `Array(unsafeUninitializedCapacity:initializingWith:)` instead of zeroing everything out first.

Sources/_CryptoExtras/RSA/RSA_boring.swift
```diff
@@ -315,6 +344,215 @@ extension BoringSSLRSAPublicKey {
...
+                    saltLength
+                )
+            }) == 1 else {
+                switch ERR_GET_REASON(CCryptoBoringSSL_ERR_get_error()) {
+                case RSA_R_DATA_TOO_LARGE_FOR_KEY_SIZE:
+                    throw _RSA.BlindSigning.Error.messageTooLong
+                default:
+                    throw CryptoKitError.internalBoringSSLError()
+                }
+            }
```
Lukasa, line 402: Can I propose factoring this section (and probably several others)) out to small helper functions? In fact, you could create a little enum-based namespace and turn each of these steps into functions in that namespace. That may make it a touch easier to follow, as this method has gotten very long.

Sources/_CryptoExtras/RSA/RSA_boring.swift
Lukasa: Oh, in fact we don't even need that: we can use `withUnsafeTemporaryAllocation` to avoid the heap allocation altogether.

Sources/_CryptoExtras/RSA/RSA_boring.swift
```diff
@@ -315,6 +344,215 @@ extension BoringSSLRSAPublicKey {
...
+            // 3. m = bytes_to_int(encoded_msg)
+            CCryptoBoringSSL_BN_bin2bn(&encodedMessage, encodedMessage.count, m)
+
+            // 4. c = is_coprime(m, n)
+            let n = CCryptoBoringSSL_RSA_get0_n(rsaPublicKey)
+            CCryptoBoringSSL_BN_gcd(gcd, m, n, bnCtx)
+            let c = CCryptoBoringSSL_BN_is_one(gcd) == 1
+
+            // 5. If c is false, raise an "invalid input" error and stop
+            guard c else {
```
Lukasa, line 413: Given the text above, it might be better to write this to exactly match: `if !c`

Sources/_CryptoExtras/RSA/RSA_boring.swift
```diff
@@ -315,6 +344,215 @@ extension BoringSSLRSAPublicKey {
...
+            }
+
+            // 10. z = (m * x) mod n
+            guard CCryptoBoringSSL_BN_mod_mul(z, m, x, n, bnCtx) == 1 else {
+                throw CryptoKitError.internalBoringSSLError()
+            }
+
+            // 11. blinded_msg = int_to_bytes(z, modulus_len)
+            var blindedMessageBytes = [UInt8](repeating: 0, count: outputSize)
+            guard CCryptoBoringSSL_BN_bn2bin_padded(&blindedMessageBytes, outputSize, z) == 1 else {
```
Lukasa, line 442: Same note here, we can use `Array(unsafeUninitializedCapacity:initializingWith:)` to avoid the all-zero initialization.

Sources/_CryptoExtras/RSA/RSA_boring.swift
```diff
@@ -315,6 +344,215 @@ extension BoringSSLRSAPublicKey {
...
+                throw CryptoKitError.internalBoringSSLError()
+            }
+
+            // 12. output blinded_msg, inv
+            let blindedMessage = Data(blindedMessageBytes)
+            var invBytes = [UInt8](repeating: 0, count: outputSize)
```
Lukasa, line 448: This is another good spot for an `unsafeTemporaryAllocation`, as that way we can avoid the temporary `Data` object.

Sources/_CryptoExtras/RSA/RSA_boring.swift
Lukasa: Oh, no, we need this as a `Data`. So a helper function, and then `withUnsafeTemporaryAllocation` and then a construction of a `Data` from the pointer.

Sources/_CryptoExtras/RSA/RSA_boring.swift
```diff
@@ -315,6 +344,215 @@ extension BoringSSLRSAPublicKey {
...
+            }
+
+            // 2. z = bytes_to_int(blind_sig)
+            try signature.rawRepresentation.withUnsafeBytes { blindSignatureBytesPtr in
+                guard CCryptoBoringSSLShims_BN_bin2bn(blindSignatureBytesPtr.baseAddress, blindSignatureBytesPtr.count, z) == z else {
+                    throw CryptoKitError.internalBoringSSLError()
+                }
+            }
+
+            // 3. s = (z * inv) mod n
```
Lukasa, line 496: Same note here: this is a long method, it would be nice to split it into multiple smaller methods.

Sources/_CryptoExtras/RSA/RSA_boring.swift
```diff
@@ -315,6 +344,215 @@ extension BoringSSLRSAPublicKey {
...
+                    throw CryptoKitError.internalBoringSSLError()
+                }
+            }
+            let n = CCryptoBoringSSL_RSA_get0_n(rsaPublicKey)
+            guard CCryptoBoringSSL_BN_mod_mul(s, z, inv, n, bnCtx) == 1 else {
+                throw CryptoKitError.internalBoringSSLError()
+            }
+
+            // 4. sig = int_to_bytes(s, modulus_len)
+            var sigBytes = [UInt8](repeating: 0, count: outputSize)
```
Lukasa, line 508: This one wants to be `withUnsafeTemporaryAllocation` as well.

Sources/_CryptoExtras/RSA/RSA_boring.swift
```diff
@@ -564,9 +814,79 @@ extension BoringSSLRSAPrivateKey {
...
+            let rsaPrivateKey = CCryptoBoringSSL_EVP_PKEY_get0_RSA(self.pointer)
+            let signatureByteCount = Int(CCryptoBoringSSL_RSA_size(rsaPrivateKey))
+
+            guard message.count == signatureByteCount else {
+                throw CryptoKitError.incorrectParameterSize
+            }
+
+            let messageBytes: ContiguousBytes = message.regions.count == 1 ? message.regions.first! : Array(message)
+
+            let signatureBytes = try Array<UInt8>(unsafeUninitializedCapacity: signatureByteCount) { signatureBufferPtr, signatureBufferCount in
```
Lukasa, line 827: I know I mentioned this transformation, but I missed that this ends up as a `Data`. We should use `withUnsafeTemporaryAllocation` here.

Sources/_CryptoExtras/Util/RandomBytes.swift
```diff
@@ -39,3 +39,15 @@ extension UnsafeMutableRawBufferPointer {
...
+
+extension SystemRandomNumberGenerator {
+    @inlinable
+    static func randomBytes(count: Int) -> [UInt8] {
+        #if canImport(Darwin) || os(Linux) || os(Android) || os(Windows)
+        var rng = SystemRandomNumberGenerator()
+        return (0..<count).map { _ in rng.next() }
+        #else
+        fatalError("No secure random number generator on this platform.")
+        #endif
```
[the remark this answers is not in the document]
simonjbeaumont: Done. I lifted this from some other places where we're generating random bytes and presumably they could be made faster too by moving over to this new function now it makes use of `Array(unsafeUninitializedCapacity:initializingWith:)`. I'll keep that out of this PR though.

Sources/_CryptoExtras/RSA/RSA+BlindSigning.swift
simonjbeaumont: I chose not to wrap these because we already have some wrapper types for the inverse and because the RFC function is returning two things.

I had a `BlindedMessage` type but this didn't work super well as an opaque type because the client needs to send just the message bytes (not the inverse, ofc) to the server. It also seemed that the result of running this protocol was to get to the point where you had a byte string and the blinding inverse.

I guess we could reinstate the `BlindedMessage` type but make it _not_ opaque:

```swift
struct BlindedMessage {
    public var blindedMessage: Data
    public var blindInverse: _RSA.BlindSigning.BlindInverse
}
```

What I don't love about this is (1) the nested repeated names and (2) I don't want API users to think they need to serialize and send this type (all of it) off to the server.

Do you have any suggestions for something clearer?

Sources/_CryptoExtras/RSA/RSA+BlindSigning.swift
simonjbeaumont: I have a local patch for this suggestion but I haven't pushed it yet because I wanted to discuss it more. What do we expect people to do with these errors? Just be able to branch on the different ones? For that, I guess we need an enum-like struct wrapper with some public lets?

Sources/_CryptoExtras/RSA/RSA+BlindSigning.swift
simonjbeaumont: Heh, I meant "should not attempt" because "cannot" but I've streamlined the language to just "cannot" here and elsewhere.

Sources/_CryptoExtras/RSA/RSA+BlindSigning.swift
simonjbeaumont: I've updated the data construction but we cannot make these `@inlinable` without some larger blast radius that would affect the existing RSA code.

Specifically because the backing key is private and there's a lot of very deliberate `private` and `fileprivate` layering which would all need to be diluted to make way for the `@usableFromInline`.

Should we defer this from this PR?

Sources/_CryptoExtras/RSA/RSA+BlindSigning.swift
simonjbeaumont: OK, I've made sure that we only throw the existing, public `CryptoKitError` through the API.

Sources/_CryptoExtras/RSA/RSA+BlindSigning.swift
Lukasa: My inclination is to call the return type `BlindingResult`, then have `blindedMessage` and `inverse`. That strikes me as the better spelling. The remaining concerns can be addressed with documentation.

Sources/_CryptoExtras/RSA/RSA_boring.swift
simonjbeaumont: OK, I've factored this and a bunch of other things into a `fileprivate enum BlindSigningHelpers`. Hopefully things should read a bit clearer now.

Tests/_CryptoExtrasTests/TestRSABlindSigning.swift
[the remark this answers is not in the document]
simonjbeaumont: This is resolved now and we don't need to use `SecureBytes` at all.

Sources/_CryptoExtras/RSA/RSA+BlindSigning.swift
simonjbeaumont: Done.

Sources/_CryptoExtras/RSA/RSA_boring.swift
Lukasa: I think we can still go further here. In particular, the unsafe temporary allocations can probably be factored into safer helper methods.

Sources/_CryptoExtras/RSA/RSA_boring.swift
Lukasa: This got thumbs-uped, but not actually done.

Sources/_CryptoExtras/RSA/RSA_boring.swift
simonjbeaumont: OK, I tried to find the right balance of keeping the steps clear from the RFC and factoring out the unsafe bits so I think the flow of this is now sensible. Specifically, if I have separate helpers for the separate steps from the RFC commentary, EEMSA-PSS-ENCODE, and bytes_to_int, then I must have a temporary allocation here.

To help converge this though, I've pushed these into one helper, abstracting over steps 1-3 which does remove this one last use of with unsafe temporary allocation from this function.

Sources/_CryptoExtras/RSA/RSA_boring.swift
simonjbeaumont: Sorry, I missed this one. Done.

simonjbeaumont:
@Lukasa as we discussed, now that we have `ArbitraryPrecisionInteger` and `FiniteFieldArithmeticContext` moved to `CryptoBoringWrapper`, I've replaced the use of manually managed `BIGNUM` in the implementation with the memory managed types. Hopefully this makes it much easier to be confident in the correctness.

simonjbeaumont:

Sources/_CryptoExtras/Util/BoringSSLHelpers.swift
```diff
@@ -93,4 +93,86 @@ extension FixedWidthInteger {
...
+    /// This constructor is used in tests in cases where test vectors provide the key information this way.
+    ///
+    /// Only the BoringSSL backend provides APIs to create the key from its parameters so we first create a BoringSSL
+    /// key, serialize it to PEM format, and then construct a platform specific key from the PEM representation.
+    internal init(nHexString: String, eHexString: String, parameters: Parameters) throws {
+        let n = try ArbitraryPrecisionInteger(hexString: nHexString)
+        let e = try ArbitraryPrecisionInteger(hexString: eHexString)
+
+        // Create BoringSSL RSA key.
+        guard let rsaPtr = (n.withUnsafeBignumPointer { n in
```
Lukasa, line 109:
```suggestion
        guard let rsaPtr = n.withUnsafeBignumPointer({ n in
```

Sources/_CryptoExtras/Util/BoringSSLHelpers.swift
```diff
@@ -93,4 +93,86 @@ extension FixedWidthInteger {
...
+
+        // Compute the CRT params.
+        let dp = try FiniteFieldArithmeticContext(fieldSize: p - 1).residue(d)
+        let dq = try FiniteFieldArithmeticContext(fieldSize: q - 1).residue(d)
+        guard let qi = try FiniteFieldArithmeticContext(fieldSize: p).inverse(q) else {
+            throw CryptoKitError.internalBoringSSLError()
+        }
+
+        // Create BoringSSL RSA key.
+        guard let rsaPtr = (n.withUnsafeBignumPointer { n in
```
Lukasa, line 150:
```suggestion
        guard let rsaPtr = n.withUnsafeBignumPointer({ n in
```

Lukasa (commented):
Generally looking really lovely, just a few quick notes here.

Sources/_CryptoExtras/RSA/RSA_boring.swift
```diff
@@ -315,6 +345,95 @@ extension BoringSSLRSAPublicKey {
...
+            // 7. inv = inverse_mod(r, n)
+            // 8. If inverse_mod fails, raise a "blinding error" error and stop
+            // NOTE: We retry here until we get an appropriate r, which is suggested.
+            let (r, inv) = try {
+                while true {
+                    let r = try ArbitraryPrecisionInteger.random(inclusiveMin: 1, exclusiveMax: n)
+                    guard let inv = try finiteField.inverse(r) else { continue }
+                    return (r, inv)
+                }
+            }()
```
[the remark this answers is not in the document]
simonjbeaumont: OK, I've updated it to use `var`, although I had this construction to ensure a `let` binding because during the last round of review, in addition to dealing with pointers, one of the comments was keeping track of which state had been computed and which was mutable.

Sources/_CryptoExtras/RSA/RSA_boring.swift
```diff
@@ -315,6 +345,95 @@ extension BoringSSLRSAPublicKey {
...
         }
 
+        fileprivate func blind<H: HashFunction>(
+            _ message: _RSA.BlindSigning.PreparedMessage,
+            parameters: _RSA.BlindSigning.Parameters<H>
+        ) throws -> _RSA.BlindSigning.BlindingResult {
+            let rsaPublicKey = CCryptoBoringSSL_EVP_PKEY_get0_RSA(self.pointer)
+            let modulusByteCount = Int(CCryptoBoringSSL_RSA_size(rsaPublicKey))
+            let e = try ArbitraryPrecisionInteger(copying: CCryptoBoringSSL_RSA_get0_e(rsaPublicKey))
+            let n = try ArbitraryPrecisionInteger(copying: CCryptoBoringSSL_RSA_get0_n(rsaPublicKey))
```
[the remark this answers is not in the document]
simonjbeaumont: I tried this out earlier and it didn't seem great because each of these computed properties would need to to use `PKEY_get0_RSA()` first, which led to a lot of duplicated calls, because we store an `EVP_PKEY *`, rather than an `RSA *` as the backing pointer.

I think these lines are pretty good for local reasoning as none of them escape a pointer beyond the rvalue used in `ArbitraryPrecisionInteger(copying:)`.

Sources/CryptoBoringWrapper/ArbitraryPrecisionInteger.swift
```diff
@@ -381,6 +408,49 @@ extension ArbitraryPrecisionInteger: SignedNumeric {
...
+        return result
+    }
+
+    @inlinable
+    public func isCoprime(with other: ArbitraryPrecisionInteger) throws -> Bool {
+        try Self.gcd(self, other) == 1
+    }
+
+    @inlinable
+    public static func random(inclusiveMin: UInt64, exclusiveMax: ArbitraryPrecisionInteger) throws -> ArbitraryPrecisionInteger {
```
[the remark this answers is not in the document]
simonjbeaumont: This one is an interesting one! I initially reached for a `Range` but a range needs to have a homogeneous type and the underlying call we're wrapping does not take two `BIGNUM*`, but instead one `BN_ULONG` and one `BIGNUM*`.

This left me with two options:

1. Write it in terms of a `Range<ArbitraryPrecisionInteger>` and fail if the lower bound is too large.
2. Write it like this.

When possible I thought we should be providing API that prohibits invalid use if we can leverage the type system to do so.

simonjbeaumont (commented):
@Lukasa addressed most of the feedback. The outstanding items that need further discussion are:

- https://github.com/apple/swift-crypto/pull/232/files#r1651172719, where I'm not sure we should provide the computed properties.
- https://github.com/apple/swift-crypto/pull/232/files#r1651153611, where a range isn't expressible without some sort of fallible API surface.

Lukasa (approved):

Sources/CryptoBoringWrapper/ArbitraryPrecisionInteger.swift
```diff
@@ -381,6 +408,49 @@ extension ArbitraryPrecisionInteger: SignedNumeric {
...
+        var result = ArbitraryPrecisionInteger()
+
+        guard result.withUnsafeMutableBignumPointer({ resultPtr in
+            exclusiveMax.withUnsafeBignumPointer { exclusiveMaxPtr in
+                CCryptoBoringSSL_BN_rand_range_ex(resultPtr, inclusiveMin, exclusiveMaxPtr)
```
finagolfin, line 444: This doesn't compile on Android armv7, a 32-bit platform, with [the following error from my Android CI](https://github.com/finagolfin/swift-android-sdk/actions/runs/9693797093/job/26750072229#step:12:444):
```
Sources/CryptoBoringWrapper/ArbitraryPrecisionInteger.swift:444:62: error: cannot convert value of type 'UInt64' to expected argument type 'BN_ULONG' (aka 'UInt32')
                CCryptoBoringSSL_BN_rand_range_ex(resultPtr, inclusiveMin, exclusiveMaxPtr)
                                                             ^
                                                             BN_ULONG(   )
```

simonjbeaumont: @finagolfin sorry about that oversight. I'm sure the fix should be pretty minimal but unfortunately I don't have a way of validating it and it's not part of our CI. Presumably the fix here is to take a `UInt` instead of a `UInt64`.

simonjbeaumont: Actually, I can reproduce this locally when building for watchOS simulator. Let me put up a patch.

simonjbeaumont: https://github.com/apple/swift-crypto/pull/242

PR #234, opened 2024-05-26
Fix RSA.Encryption.PublicKey unsafePEMRepresentation incorrect keySizeInBits 

anotheren:
In version 3.4, the `_RSA.Encryption.PublicKey` introduced a set of unsafe APIs to handle cases where `keySizeInBits` is less than 2048. However, in the implementation, the boundary check for `keySizeInBits` in `init(unsafePEMRepresentation:)` was incorrectly set as `self.keySizeInBits >= 2048`. Similar APIs indicate that it should actually be `self.keySizeInBits >= 1024`. This PR attempts to correct this issue.


Lukasa (commented):
Thanks, this is a really good catch. Are you interested in adding a quick unit test to prevent further regression?

anotheren:
Just added.

Lukasa:

PR #236, opened 2024-06-03
Move ArbitraryPrecisionInteger type to CryptoBoringWrapper module

simonjbeaumont:
### Motivation

Before this PR, the `Crypto` module had an internal `ArbitraryPrecisionInteger` type that manages the lifetime of a BoringSSL `BIGNUM` and exposes several `BIGNUM` operations. We'd like to make use of this from the `_CryptoExtras` module and also for code that will run on Darwin platforms.

### Modifications

- Move `ArbitraryPrecisionInteger` to the `CryptoBoringWrapper` module.
- Remove compilation guards so it's available on all platforms.
- Mark most operations `public`.
- Provide `RawPointer` and `MutableRawPointer` APIs (and hide typed pointer APIs).
- Provide shim in Crypto for typed `BIGNUM` pointer APIs.
- Move `ArbitraryPrecisionIntegerTests` to (new) `CryptoWrapperTests` test target.
 
There are a number of other ways this could have been achieved but none of them seemed great given the requirements, specifically:

- We require that any import of the `CCryptoBoringSSL` module is imported using `@_implementationOnly`.
- We don't want to expose `ArbitraryPrecisionInteger` (or, of course, `BIGNUM`) in the public API of `Crypto`.
- We would like to not remove the existing `@inlineable` and `@usableFromInline` annotations on existing APIs that make use of `ArbitraryPrecisionInteger`, which itself was `@usableFromInline` before this PR.

### Result

`ArbitraryPrecisionInteger` is now in a shared place that can be used from the `_CryptoExtras` module.


simonjbeaumont:
> * We require that any import of the `CCryptoBoringSSL` module is imported using `@_implementationOnly`.

@Lukasa based on our discussion, we believe that this is not a requirement and that we can actually drop the `@_implementationOnly` on our imports, which we can front-run in its own, targeted PR?

Sources/Crypto/Util/BoringSSL/ArbitraryPrecisionIntegerPointerHelpers.swift
simonjbeaumont, line 1: FYI I think we can drop the proposed raw pointer APIs on `ArbitraryPrecisionInteger` and these shims back to typed pointers if we're willing to drop `@_implementationOnly import` in this project.

simonjbeaumont:

Lukasa (approved):

Lukasa:
Thanks for this Si, much appreciated.

PR #237, opened 2024-06-06
extras: Add meaningful description to BoringSSL errors

simonjbeaumont:
### Motivation

When an error occurs in a BoringSSL call, we typically embed the packed error code in a `CryptoKitError.underlyingCoreCryptoError(error: Int32)`. This packed error code consists of the library and the reason and BoringSSL has functions for unpacking these and getting descriptive messages for them. I often spend my time in a debugger, manually calling these C functions and then looking things up in header files. 

### Modifications

In `_CryptoExtras` (not in `Crypto`) add a (possibly retroactive) conformance to `CustomStringConvertible`. 

### Result

If `_CryptoExtras` is imported, errors will likely be much more informative. For example,

```
error: "underlyingCoreCryptoError(error: 50331755)"  // before

error: "lib: bignum routines, reason: INPUT_NOT_REDUCED, code: 50331755"  // after
```

### DIFF
--- a/Sources/_CryptoExtras/Util/CryptoKitErrors_boring.swift
+++ b/Sources/_CryptoExtras/Util/CryptoKitErrors_boring.swift
@@ -22,3 +22,28 @@ extension CryptoKitError {
         return .underlyingCoreCryptoError(error: Int32(bitPattern: CCryptoBoringSSL_ERR_get_error()))
     }
 }
+
+extension CryptoKitError {
+    public var description: String {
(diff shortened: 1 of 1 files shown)


simonjbeaumont:
So, I've had this kicking around on my machine as an executable target, called `crypto-errno`. Do you think that'd be a better way to have this information provided for development and, if so, worth having in the repo?

Lukasa:
We could do both.

PR #238, opened 2024-06-11
Reduce the number of heap allocations when computing hash digests

baarde:
This PR reduces the number of heap allocations needed to compute a hash digest from 7 to 1.

### Checklist

### Motivation:

Currently, computing a hash digest requires 7 heap allocations:
1. A first `DigestContext` object is created during init.
2. Its `EVP_MD_CTX` struct is allocated by `EVP_MD_CTX_new`.
3.  `md_data` is allocated when initializing the context using `EVP_DigestInit`.
4. A second `DigestContext` object is created during finalize
2. Its `EVP_MD_CTX` struct is allocated by `EVP_MD_CTX_new`.
3.  `md_data` is allocated when copying the context using `EVP_MD_CTX_copy`.
7. A temporary array is created to store the result of `EVP_DigestFinalize`.

This means that for small messages the `HashFunction.hash` method spends most of its time allocating and deallocating memory.

Generally, users will only compute a few hash digests and aren't really concerned by the performance. But sometimes, it may be necessary to compute a lot of hash digests as fast as possible. Possible use cases include:
* Implementing the PBKDF2 key derivation function, which isn't implemented in Swift Crypto [yet](https://github.com/apple/swift-crypto/pull/98).
* Implementing [some other key derivation function](https://datatracker.ietf.org/doc/html/rfc7292#appendix-B) that definitively won't be implemented in Swift Crypto.

### Modifications:

* The temporary array is replaced by a temporary stack-allocated buffer.
* The specialized `MD5_*`/`SHA*_*` APIs are used instead of the generic `EVP_MD_*` APIs.
* The `MD5_CTX`/`SHA*_CTX` digest context is stored inline in the `DigestContext` object.
* When calling `finalize`, the temporary digest context copy is allocated on the stack as a temporary variable.
* The context and digest buffers are zeroized after use.

### Result:

On its own, this change makes hashing about 2x faster and computing HMAC authentication codes about 1.5x faster.

<details>
<summary>On my machine (MacBook Pro 14-inch 2021) with suggested changes (<code>@exclusivity(unchecked)</code> between parentheses) </summary>

MD5: 19.1 → 10.5 seconds (9.0 seconds)
SHA1: 15.8 → 7.5 seconds (6.2 seconds)
SHA256: 16.1 → 7.7 seconds (6.3 seconds)
SHA384: 20.4 → 10.6 seconds (9.2 seconds)
SHA512: 19.8 → 10.6 seconds (9.1 seconds)
HMAC-MD5: 50.4 → 31.0 seconds (28.6 seconds)
HMAC-SHA1: 42.9 → 25.1 seconds (22.9 seconds)
HMAC-SHA256: 43.5 → 25.7 seconds (23.1 seconds)
HMAC-SHA384: 52.0 → 31.4 seconds (29.2 seconds)
HMAC-SHA512: 50.5 → 32.0 seconds (29.3 seconds)
</details>
<details>
<summary>Test code</summary>

```swift
import Crypto
import Foundation

func testDigest<H: HashFunction>(_ type: H.Type) {
    let start = Date()
    let iterations = 50_000_000
    var hasher = H()
    let original = hasher
    
    var digest = hasher.finalize()
    for _ in 1 ..< iterations {
        hasher = original
        digest.withUnsafeBytes { bytes in
            hasher.update(data: bytes)
        }
        digest = hasher.finalize()
    }
    
    let duration = -start.timeIntervalSinceNow
    print("\(H.self): \(String(format: "%.1f", duration)) seconds")
}

func testHMAC<H: HashFunction>(_ type: H.Type) {
    let start = Date()
    let iterations = 50_000_000
    let key = SymmetricKey(data: Data())
    var hmac = HMAC<H>(key: key)
    let original = hmac
    
    var digest = hmac.finalize()
    for _ in 1 ..< iterations {
        hmac = original
        digest.withUnsafeBytes { bytes in
            hmac.update(data: bytes)
        }
        digest = hmac.finalize()
    }
    
    let duration = -start.timeIntervalSinceNow
    print("HMAC-\(H.self): \(String(format: "%.1f", duration)) seconds")
}

testDigest(Crypto.Insecure.MD5.self)
testDigest(Crypto.Insecure.SHA1.self)
testDigest(Crypto.SHA256.self)
testDigest(Crypto.SHA384.self)
testDigest(Crypto.SHA512.self)

testHMAC(Crypto.Insecure.MD5.self)
testHMAC(Crypto.Insecure.SHA1.self)
testHMAC(Crypto.SHA256.self)
testHMAC(Crypto.SHA384.self)
testHMAC(Crypto.SHA512.self)
```
</details>


Sources/Crypto/Digests/Digests.swift
```diff
@@ -23,14 +23,10 @@
 public struct SHA256Digest: DigestPrivate {
     let bytes: (UInt64, UInt64, UInt64, UInt64)
     
-    init?(bufferPointer: UnsafeRawBufferPointer) {
-        guard bufferPointer.count == 32 else {
-            return nil
-        }
-
+    init(initializingWith initializer: (UnsafeMutableRawBufferPointer) -> Void) {
```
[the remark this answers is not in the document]
baarde: Good idea. I'll do that.

baarde:
@Lukasa Thanks for your feedback. I've applied your suggestions. I agree: it is better.

Sources/Crypto/Digests/BoringSSL/Digest_boring.swift
```diff
@@ -56,92 +56,134 @@ struct OpenSSLDigestImpl<H: BoringSSLBackedHashFunction> {
...
 
-        static let sha384 = DigestType(CCryptoBoringSSL_EVP_sha384())
+        static let sha384 = DigestType(CCryptoBoringSSL_EVP_sha384(), SHA384_CTX.self)
 
-        static let sha512 = DigestType(CCryptoBoringSSL_EVP_sha512())
+        static let sha512 = DigestType(CCryptoBoringSSL_EVP_sha512(), SHA512_CTX.self)
+
+        fileprivate typealias SHA384_CTX = SHA512_CTX
+
+        fileprivate func withUnsafeContext<R>(data: UnsafeMutableRawPointer, _ body: (inout EVP_MD_CTX) -> R) -> R {
```
Lukasa, line 176: This code slightly awkwardly splits the baby: we store a digest specific context, and then rely on knowing how EVP_MD_CTX works internally to use it through the EVP_MD APIs, some of the time.

I don't think we should do it that way: we should either consistently use EVP_MD_CTX, or consistently use the typed digest, but not either-or.

If the goal is really to avoid the maximal number of allocations, that probably relies on us using the underlying digest contexts. That's fine, but if we're doing that we should be consistent with that, which means avoiding using the `EVP_MD_` APIs and instead using the `SHA_*` APIs. This will involve rebuilding the EVP_MD APIs, but that's not totally unreasonable if the goal is minimum number of allocations.

baarde: Thanks for the feedback. I've rewritten the code to use the base `SHA_*` APIs.

baarde:
I've rewritten the code to use the specialized `MD5_*`/`SHA*_*` APIs instead of the generic `EVP_MD_*` APIs.

I've also made sure that context and digest buffers get zeroized after use (`EVP_DigestFinal` and `EVP_MD_CTX_free` already did that for context buffers).

One noticeable "hack" is the use of the `@exclusivity` attribute to get a +15% performance improvement. This attribute was proposed in [SE-0176](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0176-enforce-exclusive-access-to-memory.md) and [implemented](https://github.com/swiftlang/swift/commit/5fd941eb677cdc97333deb051c8811a25fd77c70) in Swift 5.7, but is currently undocumented.

Sources/Crypto/Digests/BoringSSL/Digest_boring.swift
```diff
@@ -19,44 +19,118 @@ import CCryptoBoringSSL
 protocol HashFunctionImplementationDetails: HashFunction where Digest: DigestPrivate {}
 
 protocol BoringSSLBackedHashFunction: HashFunctionImplementationDetails {
-    static var digestType: DigestContext.DigestType { get }
+    associatedtype Context
+    static var digestSize: Int { get }
+    static func initialize() -> Context
+    static func update(_ context: inout Context, data: UnsafeRawBufferPointer)
+    static func finalize(_ context: inout Context, digest: UnsafeMutableRawBufferPointer)
```
Lukasa, line 26: I really like this as an API spelling, it's a lovely idea.

Sources/Crypto/Digests/BoringSSL/Digest_boring.swift
```diff
@@ -67,81 +141,44 @@ struct OpenSSLDigestImpl<H: BoringSSLBackedHashFunction> {
...
-    // This finalize function is _destructive_: do not call it if you want to reuse the object!
-    func finalize() -> [UInt8] {
-        let digestSize = CCryptoBoringSSL_EVP_MD_size(self.contextPointer.pointee.digest)
-        var digestBytes = Array(repeating: UInt8(0), count: digestSize)
-        var count = UInt32(digestSize)
-
-        digestBytes.withUnsafeMutableBufferPointer { digestPointer in
-            assert(digestPointer.count == count)
-            CCryptoBoringSSL_EVP_DigestFinal(self.contextPointer, digestPointer.baseAddress, &count)
+    func finalize<R>(_ body: (UnsafeRawBufferPointer) -> R) -> R {
```
Lukasa, line 166: I think we can avoid the extra generic here and just return `H.Digest`, as we already have the generic type parameter.

Sources/Crypto/Digests/BoringSSL/Digest_boring.swift
```diff
@@ -67,81 +141,44 @@ struct OpenSSLDigestImpl<H: BoringSSLBackedHashFunction> {
...
             // We force unwrap here because if the digest size is wrong it's an internal error.
             H.Digest(bufferPointer: $0)!
         }
     }
 }
 
-class DigestContext {
-    private var contextPointer: UnsafeMutablePointer<EVP_MD_CTX>
+final class DigestContext<H: BoringSSLBackedHashFunction> {
+    @exclusivity(unchecked) private var context: H.Context
```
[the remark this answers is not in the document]
baarde: The performance impact is not very high: about 1.15x in my use case. I'll remove the `@exclusivity` attribute.

I've done a little bit of testing. I'm able to convince the optimizer to eliminate the exclusivity checks when the following conditions are both satisfied.
1. `self.context` is not accessed as an `inout` variable.
2. `self.context` is not mutated in `deinit`.

The first condition makes sense, as the optimizer probably doesn't know that the `DigestContext` is uniquely referenced when `update` is called and must therefore assume that `CCryptoBoringSSL_*_Update` could access the context a second time using global memory.

The second condition is more puzzling to me. I don't understand how mutating the context in `deinit` could trigger an exclusivity violation, as we know for sure that `self` isn't referenced anywhere anymore.

If we decided to let go of zeroization and remove the `deinit` method, we could rewrite the `update` method like this to eliminate the exclusivity checks:
```swift
var copyContext = self.context
H.update(&copyContext, data: data)
self.context = copyContext
```
Despite the extra copies, it is faster than checking for exclusive access.

Sources/Crypto/Digests/BoringSSL/Digest_boring.swift
baarde: It's done. 🙂

Sources/Crypto/Digests/BoringSSL/Digest_boring.swift
```diff
@@ -19,44 +19,118 @@ import CCryptoBoringSSL
...
+    }
+
+    static func initialize() -> MD5_CTX {
+        var context = MD5_CTX()
+        _ = CCryptoBoringSSL_MD5_Init(&context)
+        return context
+    }
+
+    static func update(_ context: inout MD5_CTX, data: UnsafeRawBufferPointer) {
+        _ = CCryptoBoringSSL_MD5_Update(&context, data.baseAddress, data.count)
```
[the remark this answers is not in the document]
baarde: It's done. 🙂

Sources/Crypto/Digests/BoringSSL/Digest_boring.swift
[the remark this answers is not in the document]
Lukasa: Ok, I think to avoid blocking this on that performance wrinkle we should go without the exclusivity change for now. Can you file a separate Github issue that tracks eliminating that exclusivity check, ideally with a patch that we can apply that shows the two sides? I'll then see if I can work out what's happening and come up with a way to fix it.

Lukasa (approved):

PR #247, opened 2024-07-04
extras: Provide initializers for RSA keys from RSA parameters

simonjbeaumont:
## Motivation

We currently offer APIs to construct RSA keys from PEM and DER representations but we have no way of constructing they key from its constituent RSA parameters: _n_ and _e_ for public keys, and _n_, _e_, _d_, _p_, and _q_ for private keys. Sometimes these are what you have to hand, e.g. in a JSON Web Key.

## Modifications

Provide initializers for RSA keys from RSA parameters for all the RSA key types.

To implement this, we leverage the fact that we are making use of BoringSSL in `_CryptoExtras` on all platforms. We need this because there are no APIs to construct the underlying key type on Darwin platforms from these parameters. So we do this by first creating a BoringSSL key, serializing it to PEM format, and then constructing a platform-specific key from the PEM representation.

## Result

New APIs to construct RSA keys from RSA parameters.


Sources/_CryptoExtras/RSA/RSA.swift
```diff
@@ -98,6 +98,14 @@ extension _RSA.Signing {
             }
         }
 
+        /// Construct an RSA public key with the specified parameters.
+        ///
+        /// Only the BoringSSL backend provides APIs to create the key from its parameters so we first create a BoringSSL
+        /// key, serialize it to PEM format, and then construct a platform specific key from the PEM representation.
+        public init<D: ContiguousBytes>(n: D, e: D) throws {
+            try self.init(pemRepresentation: BoringSSLRSAPublicKey(n: n, e: e).pemRepresentation)
```
Lukasa, line 106: Let's have a backing implementation that always accepts a BoringSSL key, and then implement it twice. This avoids the encode/decode on Linux, where we don't have to pay for it.

Sources/_CryptoExtras/RSA/RSA.swift
```diff
@@ -98,6 +98,14 @@ extension _RSA.Signing {
             }
         }
 
+        /// Construct an RSA public key with the specified parameters.
+        ///
+        /// Only the BoringSSL backend provides APIs to create the key from its parameters so we first create a BoringSSL
+        /// key, serialize it to PEM format, and then construct a platform specific key from the PEM representation.
+        public init<D: ContiguousBytes>(n: D, e: D) throws {
```
Lukasa, line 105: All of these need multiple generic type parameters. There's no particularly good reason to force each parameter to be the same type.

simonjbeaumont: Yep. I was thinking the same thing. Originally I had these take hex strings but I think bytes is probably better. Will make this more flexible 👍

Sources/_CryptoExtras/RSA/RSA.swift
simonjbeaumont: That’s a good idea. Alternatively we could `#if` dance in here but I didn’t want it to get out of step with the compiler conditions elsewhere. I’ll do the backing impl. Maybe that will motivate moving some of the code _back_ into BoringSSLHelpers so it can be shared by both backing implementations?

Sources/_CryptoExtras/RSA/RSA.swift
simonjbeaumont: Updated so that each generic parameter can be a different concrete type conforming to `ContiguousBytes` and added a test that would have failed before, but now passes.

Sources/_CryptoExtras/RSA/RSA.swift
simonjbeaumont: I added the backing initializers that accept the BoringSSL key and now we don't have the conversion on BoringSSL-backed platforms.

simonjbeaumont (commented):
Thanks for the feedback @Lukasa. I think this is ready for another pass 🙏

Sources/_CryptoExtras/RSA/RSA_security.swift
```diff
@@ -43,6 +43,10 @@ internal struct SecurityRSAPublicKey: @unchecked Sendable {
         self.backing = unwrappedKey
     }
 
+    init(_ boringSSLKey: BoringSSLRSAPublicKey) throws {
+        try self.init(pemRepresentation: boringSSLKey.pemRepresentation)
```
Lukasa, line 47: It's probably a touch cheaper to use the DER representation.

Lukasa (commented):
Very nice, just a little note.

0xTim:
cc @ptoffy since we've implemented that ourselves in JWTKit - we should either wait for this or ensure our APIs are internal so we can switch over without an API break

ptoffy:
@0xTim we have our own key wrapping SwiftCrypto's so we should be fine. We can switch to this without breaking anything

Sources/_CryptoExtras/RSA/RSA_security.swift
simonjbeaumont: Thanks, done!

Sources/_CryptoExtras/RSA/RSA.swift
```diff
@@ -98,6 +98,17 @@ extension _RSA.Signing {
...
         }
 
+        /// Construct an RSA public key with the specified parameters.
+        ///
+        /// Only the BoringSSL backend provides APIs to create a key from its parameters so we first create a BoringSSL
+        /// key, and then pass it to the platform-specific initializer that accepts a BoringSSL key.
+        ///
+        /// On Darwin platforms, this will serialize it to PEM format, and then construct a platform-specific key from
+        /// the PEM representation.
+        public init(n: some ContiguousBytes, e: some ContiguousBytes) throws {
```
Lukasa, line 108: Generalised nit: `@inlinable` on generic methods, as far as possible. This feedback applies to any new generic function or type.

Lukasa (commented):
Great, looks good to me, minus the inlinable thing.

Sources/_CryptoExtras/RSA/RSA.swift
simonjbeaumont: FWICT there is a very deliberate use of `private` for the backing key types and `fileprivate` for  the type aliases which is why _existing_ generic initialisers are not currently inlinable either (e.g. the PEM and DER initialisers).

I think we'd probably need to take a wider view to achieve this so should we defer this to a future PR?

simonjbeaumont:

Sources/_CryptoExtras/RSA/RSA.swift
Lukasa: Sounds good.

Lukasa (approved):

PR #249, opened 2024-08-08
extras: Add RSA primitives decoding API

ptoffy:
### Motivation:

At times we might need to get the raw elements of an RSA key, e.g when debugging.

### Modifications:

Add a `getKeyPrimitives` method to the RSA (public and private) key structures that returns the elements.
To do this we leverage the built in BoringSSL methods (`CCryptoBoringSSL_RSA_get0_n` etc.) which return the raw elements of the key. In the Security version of the key we convert the key to a BoringSSL one and just call the BoringSSL method. 

Once we get ASN.1 support in `_CryptoExtras` it might make sense switching the implementation of this method to decode the ASN.1 structure and returns the elements from there.

### Result:

New APIs to fetch the key primitives.


Lukasa:
Thanks for this!

Can you clarify _when_ exactly you want to do this? When debugging is not a strong argument in my view, as we could offer a printable representation instead, or just use the existing serialization APIs.

As a sidebar, if we do down this road we shouldn't return tuples but instead nominal types, and we should at the very least store the bytes in a `Data`, if not a custom type that wraps `SecureBytes` for safety reasons.

ptoffy:
So we got a request for this in JWTKit (https://github.com/vapor/jwt-kit/pull/77) and implemented it when we were still using BoringSSL. When removing BoringSSL we wanted to keep the API so we made our own ASN.1 decoding but I thought I might try adding it to Crypto so we can remove one more dependency

Lukasa:
Gotcha. I think this API is fine to have _if_ we try to make it a bit safer to use, so let's restrict ourselves to only doing this on the public keys. I believe that will resolve the use-case you had here.

Sources/_CryptoExtras/RSA/RSA+BlindSigning.swift
```diff
@@ -113,6 +113,16 @@ extension _RSA.BlindSigning {
             self.backing = backing
             self.parameters = parameters
         }
+
+        public struct Primitives {
+            public var n: Data
+            public var e: Data
```
Lukasa, line 119: For this type, can we use the long names for these fields? `modulus` and `public exponent`. Let's also make this type `Hashable`, `Sendable`, and give it a `public init`. Same for all the others.

Sources/_CryptoExtras/RSA/RSA_boring.swift
```diff
@@ -465,6 +469,22 @@ extension BoringSSLRSAPublicKey {
...
             CCryptoBoringSSL_EVP_PKEY_free(self.pointer)
         }
+
+        fileprivate func getKeyPrimitives() throws -> (n: Data, e: Data) {
+            let key = CCryptoBoringSSL_EVP_PKEY_get0_RSA(self.pointer)
+
+            func getPrimitive(_ getPointer: (OpaquePointer?) -> UnsafePointer<BIGNUM>?) throws -> Data {
+                let ptr = getPointer(key)
+                let size = Int(CCryptoBoringSSL_BN_num_bytes(ptr))
+                var buffer = [UInt8](repeating: 0, count: size)
```
Lukasa, line 479: I don't think we need to involve a poor innocent temporary Array here: we may as well compute this directly on `Data`.

Lukasa (commented):
Thanks! A few quick notes

Sources/_CryptoExtras/RSA/RSA+BlindSigning.swift
ptoffy: I used `n` and `e` to be consistent with the `init(n:, e:)` initialiser but happy to change it

Lukasa: Yeah, I regret having done that that way, but too late now!

Sources/_CryptoExtras/RSA/RSA.swift
```diff
@@ -132,6 +142,11 @@ extension _RSA.Signing {
         fileprivate init(_ backing: BackingPublicKey) {
             self.backing = backing
         }
+
+        public func getKeyPrimitives() throws -> Primitives {
```
Lukasa, line 146: I'm inclined to want to make these effectful computed properties. What do you think @ptoffy?

ptoffy: I'm honestly a bit torn 😅 having `key.modulus` and `key.publicExponent` does look cleaner, but if you mean having it as `key.primitives` we might as well leave it as a function: it throws, there's a conversion (on Darwin) and we _are_ calling BoringSSL so there _is_ some underlying computation. Odds are it won't get accessed often so caching doesn't really make sense and I'd say it's unlikely someone's accessing either one without the other(?)

Sources/_CryptoExtras/RSA/RSA_boring.swift
```diff
@@ -465,6 +469,22 @@ extension BoringSSLRSAPublicKey {
...
+
+        fileprivate func getKeyPrimitives() -> (n: Data, e: Data) {
+            let key = CCryptoBoringSSL_EVP_PKEY_get0_RSA(self.pointer)
+
+            func getPrimitive(_ getPointer: (OpaquePointer?) -> UnsafePointer<BIGNUM>?) -> Data {
+                let ptr = getPointer(key)
+                let size = Int(CCryptoBoringSSL_BN_num_bytes(ptr))
+                var data = Data(count: size)
+                data.withUnsafeMutableBytes { dataPtr in
+                    _ = CCryptoBoringSSL_BN_bn2bin(ptr, dataPtr.baseAddress!.assumingMemoryBound(to: UInt8.self))
```
Lukasa, line 481: I don't think we should need the assuming memory bound here, we should be safely able to pass the raw pointer through.

Lukasa (approved):
Great, thanks for this @ptoffy, I think this is good to go!

Issue #248, opened 2024-07-22
RSA initialiser with primes generation


ptoffy:
#### Motivation:

After #247 lands we will have an API which allows us to create RSA keys directly from the raw elements, which is great. However I think an important step would be to also add an initialiser which generates RSA keys instead of just "parsing" them, meaning that it could leverage BoringSSL to create `p` and `q` itself without relying on the user to do so. The functionality is already inside of BoringSSL (https://boringssl.googlesource.com/boringssl/+/master/crypto/fipsmodule/bn/prime.c) so we could probably just bridge to that. 

#### Importance:

Without such API users of the library could create a SwiftCrypto RSA key, though they would still have to generate the primes used in the key themselves. This means that users would have to add a dependency on an arbitrary precision library, then write prime generation algorithms (hard to get right, even harder to get performant) and just then can they use the new SwiftCrypto APIs. JWTKit is doing [so](https://github.com/vapor/jwt-kit/blob/main/Sources/JWTKit/RSA/RSA%2BKeyCalculation.swift#L7) at the moment and it works, though it's  obviously not nearly as optimised as it could/would be if we were using BoringSSL directly.
After #247 we can remove a ton of the serialisation code for RSA keys from JWTKit and if this API were to be added too we could remove a whole lot of unoptimised code (and even a dependency). This would then also be mostly inline with other keys like ECDSA etc.

I'd be open to making a PR myself but thought I'd open an issue first to test the waters

Lukasa:
Don't you want this function?

https://github.com/apple/swift-crypto/blob/8dafe0fdce623f65178689f2d6dea7304f7fbe75/Sources/_CryptoExtras/RSA/RSA.swift#L207

ptoffy:
No, I want to be able to generate an RSA key based on `n`, `e` and `d`, such as https://github.com/vapor/jwt-kit/blob/main/Sources/JWTKit/RSA/RSA%2BKeyCalculation.swift#L7

Lukasa:
Ah, gotcha. @simonjbeaumont I'd be happy with us adding this API surface as well, if you're happy to do so.

simonjbeaumont:
Sure. I can pick this up at some point.

ptoffy:
[...]

PR #250, opened 2024-08-12
extras: Add API to create RSA keys with random primes 

ptoffy:
Tied to #248 
This adds a `_createFromNumbers` API which creates an RSA key generating `p` and `q` primes based on `n`, `e` and `d`

### Motivation:

There might be cases where private RSA keys have to be generated based on their raw elements (modulus `n`, exponent `e` and private exponent `d`) without `p` and `q`. These are random primes which are integral parts of the key but can be generated based on the former numbers, and are not _always_ included when representing RSA keys, for example these parameters are not required in JWKs

### Modifications:

This adds a static `_createFromNumbers` method which creates a new RSA key given its raw modulus, exponent and private exponent, generating the needed `p` and `q` primes without having to provide them as parameters. The algorithm to generate primes is based on a hybrid approach of the Miller Rabin primality test and the proof of Fact 1 in Dan Boneh's [Twenty Years of Attacks on the RSA Cryptosystem paper](https://www.ams.org/notices/199902/boneh.pdf). To write this algorithm, some additions had to be made to the `ArbitraryPrecisionInteger` 

### Result:

Private RSA keys can now be generated without providing `p` and `q`


Sources/_CryptoExtras/Util/Error.swift
```diff
@@ -14,4 +14,5 @@
 
 public enum _CryptoRSAError: Error {
     case invalidPEMDocument
+    case keyInitializationFailure
```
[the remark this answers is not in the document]
ptoffy: Cool I'm fine with that, although wouldn't it make sense to hide the enum behind a struct so we can add/remove cases without breaking the API?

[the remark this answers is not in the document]
Lukasa: We can do that too, but to be clear I was suggesting we use an existing _case_ from `CryptoKitError`. In this instance I'm just not sure we need a specific new error.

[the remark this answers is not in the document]
ptoffy: Yeah I got that, I'm just unsure if there's a fitting one

Sources/Crypto/Util/BoringSSL/Shared/ArbitraryPrecisionInteger_boring.swift
```diff
@@ -443,6 +450,46 @@ extension ArbitraryPrecisionInteger {
...
+                }
+            }
+        }
+        precondition(rc == 1, "Unable to allocate memory for new ArbitraryPrecisionInteger")
+
+        return result
+    }
+
+    @usableFromInline
+    func isEven() -> Bool {
```
Lukasa, line 488: Nit: this should probably be a computed property, not a function.

Sources/_CryptoExtras/RSA/RSA+BlindSigning.swift
```diff
@@ -227,6 +227,21 @@ extension _RSA.BlindSigning {
         public var publicKey: _RSA.BlindSigning.PublicKey<H> {
             _RSA.BlindSigning.PublicKey(self.backing.publicKey, self.parameters)
         }
+
+        public static func _createFromNumbers(n: some ContiguousBytes, e: some ContiguousBytes, d: some ContiguousBytes, parameters: Parameters) throws -> Self {
```
Lukasa, line 231: Let's put a doc comment on here explaining what these do, along with the rest. This doc comment should make it clear that this constructor is strongly disfavoured as it's got extremely poor performance characteristics, and that it's provided as a convenience for when there is no better choice.

Sources/_CryptoExtras/RSA/RSA.swift
```diff
@@ -639,3 +667,59 @@ extension _RSA {
 
     static let SPKIPublicKeyType = "PUBLIC KEY"
 }
+
+extension _RSA {
+    static func extractPrimeFactors(
+        n: ArbitraryPrecisionInteger, 
+        e: ArbitraryPrecisionInteger, 
+        d: ArbitraryPrecisionInteger
+    ) throws -> (p: ArbitraryPrecisionInteger, q: ArbitraryPrecisionInteger) {
```
Lukasa, line 676: Can you add the paper reference as a comment here? It'll help future readers.

Tests/_CryptoExtrasTests/TestRSABlindSigning.swift
```diff
@@ -202,4 +202,26 @@ final class TestRSABlindSigning: XCTestCase {
             parameters: .RSABSSA_SHA384_PSS_Randomized
         )
     }
+
+    func testConstructAndUseKeyFromRSANumbersWhileRecoveringPrimes() throws {
+        let data = Array("hello, world!".utf8)
+
+        // Generating primes is probabilistic, so we run this test multiple times.
```
Lukasa, line 209: I'm a bit nervous about a test that _fails_ probabilistically under correct scenarios, which is what this test appears to do. We should presumably make an estimate of what the failure rate should be, scaled by the number of test iterations here, and then move the threshold of failures to two std dev.

Lukasa:

Tests/_CryptoExtrasTests/TestRSABlindSigning.swift
ptoffy: Actually I added this because with the first commit of the algorithm it was failing something like once every 5 tries. This was however due to me adding a wrong exponentiation which is now resolved. I've run this test locally a handful of times (even with 1000 iterations) and it always succeeded, hence I'm actually quite confident the algorithm works fine and we can remove both the comment and the loop

Sources/Crypto/Util/BoringSSL/Shared/ArbitraryPrecisionInteger_boring.swift
```diff
@@ -405,6 +405,13 @@ extension ArbitraryPrecisionInteger: SignedNumeric {
 // MARK: - Other arithmetic operations
 
 extension ArbitraryPrecisionInteger {
+    @usableFromInline
+    var trailingZeroBitCount: Int32 {
+        self._backing.withUnsafeBignumPointer {
```
Lukasa, line 410:
```suggestion
        self.withUnsafeBignumPointer {
```

Sources/Crypto/Util/BoringSSL/Shared/ArbitraryPrecisionInteger_boring.swift
```diff
@@ -443,6 +450,46 @@ extension ArbitraryPrecisionInteger {
...
+            }
+        }
+        precondition(rc == 1, "Unable to allocate memory for new ArbitraryPrecisionInteger")
+
+        return result
+    }
+
+    @usableFromInline
+    var isEven: Bool {
+        self._backing.withUnsafeBignumPointer {
```
Lukasa, line 489:
```suggestion
        self.withUnsafeBignumPointer {
```

Lukasa (commented):
Sorry for the delay on this, was a busy week last week. I think I have two small nits and then we're good to go.

Tests/_CryptoExtrasTests/TestRSABlindSigning.swift
Lukasa: Let's remove the loop.

ptoffy: Done!

Lukasa (approved):
Fantastic, great work @ptoffy!

PR #251, opened 2024-08-12
Convert PrivacyInfo.xcprivacy symlinks to files

jakepetroules:
It is expected behavior of the swift-build build system to copy links as-is, so presenting these as symlinks results in broken symlinks being present in the produced resource bundles, which won't work properly.

On Windows, it even causes the build entirely to fail due to a related bug: https://github.com/apple/swift-foundation/issues/845

This is a regression from #226. Convert these to flat files instead.


jakepetroules:
@swift-ci please test

Lukasa:

Lukasa (approved):
Thanks Jake

PR #252, opened 2024-08-29
Add support for AES CFB mode in _CryptoExtras

simonjbeaumont:
## Motivation

We have AES support in `_CryptoExtras` for CBC, CTR, and GCM-SIV modes; but we do not have support for CFB.

## Modifications

- Add support for AES-CFB, backed by BoringSSL, with the same shape of API as we provide for AES CTR.
- Copy RFCVector decoder from Crypto tests to CryptoExtrasTests.
- Add test vectors from NIST.

## Result

New API added to support AES-CFB.


Tests/_CryptoExtrasTests/Utils/RFCVector.swift
simonjbeaumont, line 1: Note to reviewer: This file has been copied from `CryptoTests`.

Sources/_CryptoExtras/AES/AES_CFB.swift
```diff
@@ -0,0 +1,71 @@
...
+            nonce: AES._CFB.Nonce
+        ) throws -> Data {
+            let bytes: ContiguousBytes = ciphertext.regions.count == 1 ? ciphertext.regions.first! : Array(ciphertext)
+            return try AESCFBImpl.encryptOrDecrypt(.decrypt, bytes, using: key, nonce: nonce)
+        }
+    }
+}
+
+extension AES._CFB {
+    public struct Nonce: Sendable {
```
simonjbeaumont, line 46: FYI: Our existing AES APIs differ in calling this `IV` or `Nonce`. I aligned with CTR, which was added most recently(?).

Lukasa: For CFB nonce is acceptable, though it is probably more commonly called an IV. NIST calls it an IV, though that seems more to be out of caution than anything else.

Lukasa (approved):
This looks great, thanks!

Sources/_CryptoExtras/AES/AES_CFB.swift
simonjbeaumont: Renamed `AES._CFB.Nonce` to `IV` and all `nonce` and `nonceBytes` parameters to `iv` and `ivBytes, respectively.

simonjbeaumont:
Thanks @Lukasa I just push an update that renamed `Nonce` to `IV`, throughout. Otherwise no changes since your review. I'll leave it with you to merge whenever it's a good time.

PR #255, opened 2024-09-13
Update BoringSSL to 6a2ccdcc2ed1d37a43a2183658d2ae61fd5ce208

Lukasa:
There have been a few changes in the BoringSSL build, so those are present in this commit too. Otherwise, this updates BoringSSL to 6a2ccdcc2ed1d37a43a2183658d2ae61fd5ce208.


simonjbeaumont (approved):
Changes to the vendor script look good. Didn't review every line of vendored code of course. Looks like the API break pipeline is just because the branch is outdated with main.

Sources/CCryptoBoringSSL/CMakeLists.txt
```diff
@@ -319,80 +252,21 @@ add_library(CCryptoBoringSSL STATIC
   "crypto/x509/x_spki.c"
   "crypto/x509/x_val.c"
   "crypto/x509/x_x509.c"
-  "crypto/x509/x_x509a.c")
+  "crypto/x509/x_x509a.c"
+  "gen/crypto/err_data.c")
 
 if(CMAKE_SYSTEM_NAME STREQUAL Darwin AND CMAKE_SYSTEM_PROCESSOR MATCHES "amd64|x86_64")
   target_sources(CCryptoBoringSSL PRIVATE
-    crypto/chacha/chacha-x86_64-mac.mac.x86_64.S
```
glbrntt, line 326: What happened to all of the asm? And is this big `if` needed anymore?

Lukasa: Good catch. It's still there, but I suspect the script is looking in the wrong place.

Lukasa: Ah yeah, so these are now generated by BoringSSL, and it expects the names to match.

Lukasa: Now fixed.

glbrntt (approved):
LGTM

Sources/CCryptoBoringSSL/CMakeLists.txt
glbrntt: OOI what happens if they are missing? Presumably there's a lower-perf fallback?

Lukasa: No, I think the compile would fail. You need `NO_ASM` set to fallback. We don't build the cmake version in CI here which is why we didn't notice.

glbrntt: There is a `cmake` CI job which did pass:
https://ci.swiftserver.group/job/swift-crypto-cmake-prb/306/ 

It builds the project with ninja: https://github.com/apple/swift-crypto/blob/81bee98e706aee68d39ed5996db069ef2b313d62/docker/docker-compose.yaml#L37

and it looks like we only set `NO_ASM` when building was WASI:
https://github.com/apple/swift-crypto/blob/81bee98e706aee68d39ed5996db069ef2b313d62/Package.swift#L104

This is a little surprising given the above...

Lukasa: Yeah, that's _very_ curious.

PR #260, opened 2024-09-19
Making ivBytes accessible for _CBC.IV

maschall:
### Checklist

### Motivation:

Right now, we're using CommonCrypto and would like to switch to Crypto instead.  However, the requirement we need is that we send the data for IV to the server for them to verify the signature we are sending them.  Similarly, they send us their IV and we verify the signature of other data. (Which is supported currently)

### Modifications:

By making `ivBytes` public, we can access the randomly generated salt for our cipher.

### Result:

I can replace CommonCrypto code in my repo with Crypto instead :)

### DIFF
--- a/Sources/_CryptoExtras/AES/AES_CBC.swift
+++ b/Sources/_CryptoExtras/AES/AES_CBC.swift
@@ -145,7 +145,7 @@ extension AES._CBC {
     /// An initialization vector.
     public struct IV: Sendable {
         // AES CBC uses a 128-bit IV.
-        var ivBytes: (
+        public var ivBytes: (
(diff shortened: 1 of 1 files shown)


Lukasa:
Thanks for this, this is a change we're generally happy to make. Can I make a few requests?

Firstly, let's prefer to offer an API that exposes an `Array`. That will generally be more useful than the tuple we have here.

Secondly, can you add a unit test that drives that API? It'll help us avoid regressing it.

maschall:
That is interesting about the [UInt8] because that is how I use it in my code, and had to repeat a lot of ivBytes.0, 1, 2,...

So I'm happy to expose and API and unit tests

maschall:
Closing in favor of #261

PR #261, opened 2024-09-21
Using GCM Nonce pattern for CBC, CFB, and CTR

maschall:
Following GCM use of gyb for Nonce generation, I converted IV and Nonce classes of the CBC, CFB, and CTR to allow conformance to ContinuousBytes and Sequence.

### Checklist
- [?] I've updated the documentation if necessary (Probably need to do this)

### Motivation:

The project I am working on uses a AES CBC algorithm to send encrypted data to exchange keys between a server and client.  This means, that the client and server exchange the IV/Nonce with each other.  Currently, we can only use CommonCrypto, because AES._CBC.IV doesn't facilitate an API to get the internal data bytes.

### Modifications:

Removed the current IV and Nonce structs from their corresponding files.  Made a Nonces.swift.gyb file that will generate the corresponding structs for each algorithm.

### Result:

Allows us to integrate Crypto into our repository and fix an issue we have involving padding of our cipher data, when using CCCrypt

I'm happy to add unit tests for this PR, however I don't think they are needed.  While the conformance to new Protocols is public, I believe their use is also internal and so they are being exercised by the current set of unit tests.


Lukasa (commented):
This is a really nice change in terms of getting the API laid out! However, it'd be really nice to keep the Nonces and IVs stored directly, instead of using `Data` objects. Can we return to that representation?

maschall:
Yeah, I'll give it a shot

maschall:
changed from Data back to the Tuple style.  I did give them typealiases

maschall:
Just a friendly bump: @Lukasa in case it gets lost in the shuffle

Sources/_CryptoExtras/AES/Nonces.swift
```diff
@@ -0,0 +1,298 @@
...
+        typealias IVTuple = (
+            UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, 
+            UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8
+        )
+    
+        private var bytes: IVTuple
+
+        /// Creates a new random nonce.
+        public init() {
+            var bytes = IVTuple(0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0)
```
Lukasa, line 38: I suspect the missing spaces after commas here will trigger the format checker.

Sources/_CryptoExtras/AES/Block Function.swift
```diff
@@ -131,9 +131,10 @@ extension AES {
         init(_ blockBytes: BlockBytes) {
             self.blockBytes = blockBytes
         }
-
-        init(_ iv: AES._CBC.IV) {
-            self.blockBytes = iv.ivBytes
+        
+        init<BlockBytes: Sequence>(blockBytes: BlockBytes) where BlockBytes.Element == UInt8 {
+            let blockBytes: [UInt8] = Array(blockBytes)
+            self.init(blockBytes: blockBytes)
```
Lukasa, line 137: I get the impression we can keep the existing constructor here, as the tuple is still present now.

Lukasa:

Sources/_CryptoExtras/AES/Nonces.swift
maschall: :doh:  I'll get that

maschall: I've updated this, and made a slight tweak to add a "blankBytes" variable to not have to copy the 0's in a few different places.

Sources/_CryptoExtras/AES/Block Function.swift
maschall: I added the original constructor back in

Sources/_CryptoExtras/AES/AES_CTR.swift
```diff
@@ -42,41 +43,3 @@ extension AES {
...
-            var rng = SystemRandomNumberGenerator()
-            self.nonceBytes = (
-                rng.next(), rng.next(), rng.next()
-            )
-        }
-
-        public init<NonceBytes: Collection>(nonceBytes: NonceBytes) throws where NonceBytes.Element == UInt8 {
-            // We support a 96-bit nonce (with a 32-bit counter, initialized to 0) or a full 128-bit
-            // expression.
-            guard nonceBytes.count == 12 || nonceBytes.count == 16 else {
```
maschall, line 65: I missed this part of CTR, I'll address it

Sources/_CryptoExtras/AES/Nonces.swift
```diff
@@ -0,0 +1,304 @@
...
+            return try Swift.withUnsafeBytes(of: &bytes, body)
+        }
+        
+        mutating func withUnsafeMutableBytes<ReturnType>(_ body: (UnsafeMutableRawBufferPointer) throws -> ReturnType) rethrows -> ReturnType {
+            var bytes = self.bytes
+            return try Swift.withUnsafeMutableBytes(of: &bytes, body)
+        }
+        
+        /// Returns an iterator over the elements of the nonce.
+        public func makeIterator() -> Array<UInt8>.Iterator {
```
Lukasa, line 298: Let's use our own iterator type to hide the face that we're using Array. We _can_ make this cheaper in future, and it would probably be nice to do so.

maschall: I gave it a shot, and made a generic Iterator that could be used by all three.

It might make sense to bring it into the GBC and ChaChaPoly nonces too

Sources/_CryptoExtras/AES/Nonces.swift
```diff
@@ -0,0 +1,322 @@
...
+import Crypto
+import Foundation
+// MARK: - Generated file, do NOT edit
+// any edits of this file WILL be overwritten and thus discarded
+// see section `gyb` in `README` for details.
+
+fileprivate struct ByteIterator<T>: IteratorProtocol {
+    var currentOffset = 0
+    var pointer: UnsafeRawBufferPointer? = nil
+    let length: Int
```
Lukasa, line 24: Great change! Let's drop the pointer though. In the fullness of time this may end up using `Span`, but for now we can still use the Array iterator in here. I don't think we want a pointer, so we'd have to use some other spelling.

maschall: OK.  I wasn't sure if you were suggesting to also drop the offset and length. (I was trying to figure out what you meant by `Span`)

Lukasa: Oh sorry, span is [here](https://forums.swift.org/t/pitch-2-safe-access-to-contiguous-storage/72745)

Sources/_CryptoExtras/AES/Nonces.swift.gyb
```diff
@@ -0,0 +1,155 @@
...
+            return try Swift.withUnsafeBytes(of: &bytes, body)
+        }
+        
+        mutating func withUnsafeMutableBytes<ReturnType>(_ body: (UnsafeMutableRawBufferPointer) throws -> ReturnType) rethrows -> ReturnType {
+            var bytes = self.bytes
+            return try Swift.withUnsafeMutableBytes(of: &bytes, body)
+        }
+        
+        /// Returns an iterator over the elements of the nonce.
+        public func makeIterator() -> some IteratorProtocol<UInt8> {
```
Lukasa, line 150: Let's add `@inlinable` annotations to every function on this type. It'll require `@usableFromInline` on the typealias, `var bytes`, and `static var emptyBytes` but that should be fine too.

PR #265, opened 2024-09-30
Refactor and adopt package access to remove symlinked files

simonjbeaumont:
## Motivation

Back in 3.5.0 we needed to make use of `ArbitraryPrecisionInteger` and `FiniteFieldArithmeticContext` from `_CryptoExtras`. These types lived in `Crypto` and used `@_implementationOnly import CCryptoBoringSSL`. We mistakenly removed the `@_implementationOnly` from the import after some failed analyses, resulting in compilation failures in downstream projects that had other vendored copies of BoringSSL (aside: this was in spite of the namespaced symbol rewriting which wasn't catching some BoringSSL C macro expansions).

We reverted to using `@_implementationOnly import` in 3.5.1 and resorted to using symlinks to share the code between `Crypto` and `_CryptoExtras`. This was the best we could do at the time because we still supported Swift 5.8 and so could not make use of `package` access. It also required an additional workaround define, `MODULE_IS_CRYPTO`.

We have now dropped support for 5.8 and have the opportunity to unwind this workaround.

For additional context, note that these types are needed by both:

1. (Conditionally) `Crypto`, when compiling for non-Darwin platforms; and
2. (Unconditionally) `_CryptoExtras`, when compiling features that do not have the required primitives exposed in Darwin APIs.

Fortunately, we have the internal `CryptoBoringWrapper` target, which is the existing home for exactly such code.

Finally, note that existing uses of `CryptoBoringWrapper` are importing it _with_ `@_implementationOnly` but this seems to be unnecessary since `CryptoBoringWrapper` makes use of `@_implementationOnly` when importing `CCryptoBoringSSL` (as it should).

## Modifications

- Delete the symlinks in `_CryptoExtras` to `ArbitraryPrecisionInteger.swift` and `FiniteFieldArithmeticContext.swift`.
- Remove the `MODULE_IS_CRYPTO` workaround compile-time define.
- Move `ArbitraryPrecisionInteger.swift` and `FiniteFieldArithmeticContext.swift` to `CryptoBoringWrapper`.
- Adopt `package` access for the types and all their (internal) API surface.
- Mark `ArbitraryPrecisionInteger.BackingStorage` as `fileprivate` for good measure.
- Flesh out `CryptoBoringWrapperError` to match `CryptoError` because some of these errors are used.
- Drop `@_implementationOnly` on all _existing_ imports of `CryptoBoringWrapper`.
- Import `CryptoBoringWrapper` in all places that required `ArbitraryPrecisionInteger` and/or `FiniteFieldArithmeticContext`.
- Move `ArbitraryPrecisionIntegerTests.swift` and `FiniteFieldArithmeticContextTests.swift` to new `CryptoBoringWrapperTests` test target.
- Update CMake files to account for the moved sources and to add `-package-name` to compiler arguments.

## Additional notes

Note that I tried to make use of `package import CryptoBoringWrapper` but at this time this isn't working well with `@inlinable` or `@usableFromInline` symbols.

## Results

- No adopter-visible change.
- The symlinks and manifest hacks are removed and we can continue to share code between `Crypto` and `_CryptoExtras`.

## Testing

The breakage caused by the 3.5.0 was caught by a Dependabot update on penny-bot[^0] and so I have used that project to validate that this change is safe:

- Using penny-bot commit `e9b7340` (commit at time of last break).
- Using `swift:5.10-noble` image (CI used by the previously failing pipeline).
- Using local package dependency of swift-crypto; otherwise unmodified penny-bot.
- Builds with swift-crypto 3.4.0
- Fails with swift-crypto 3.5.0 (repro of leaking BoringSSL).
- Builds with this PR patch.

[^0]: https://github.com/vapor/penny-bot/actions/runs/9767425786/job/26962669715?pr=211

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,10 @@ endif()
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 set(CMAKE_Swift_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/swift)
 
+set(SWIFT_CRYPTO_COMPILE_OPTIONS
+    -package-name com.apple.swift-crypto
+)
+
 option(BUILD_SHARED_LIBS "Build shared libraries by default" YES)
 
 if(BUILD_SHARED_LIBS)
--- a/Package.swift
+++ b/Package.swift
@@ -135,7 +135,7 @@ let package = Package(
                 "Signatures/ECDSA.swift.gyb",
             ],
             resources: privacyManifestResource,
-            swiftSettings: swiftSettings + [.define("MODULE_IS_CRYPTO")]
+            swiftSettings: swiftSettings
         ),
         .target(
             name: "_CryptoExtras",
@@ -173,6 +173,7 @@ let package = Package(
             swiftSettings: swiftSettings
         ),
         .testTarget(name: "_CryptoExtrasTests", dependencies: ["_CryptoExtras"]),
+        .testTarget(name: "CryptoBoringWrapperTests", dependencies: ["CryptoBoringWrapper"]),
     ],
     cxxLanguageStandard: .cxx11
 )
--- a/Sources/Crypto/AEADs/AES/GCM/BoringSSL/AES-GCM_boring.swift
+++ b/Sources/Crypto/AEADs/AES/GCM/BoringSSL/AES-GCM_boring.swift
@@ -15,7 +15,7 @@
 @_exported import CryptoKit
 #else
 @_implementationOnly import CCryptoBoringSSL
-@_implementationOnly import CryptoBoringWrapper
+import CryptoBoringWrapper
 import Foundation
 
 enum OpenSSLAESGCMImpl {
--- a/Sources/Crypto/AEADs/ChachaPoly/BoringSSL/ChaChaPoly_boring.swift
+++ b/Sources/Crypto/AEADs/ChachaPoly/BoringSSL/ChaChaPoly_boring.swift
@@ -16,7 +16,7 @@
 #else
 @_implementationOnly import CCryptoBoringSSL
 @_implementationOnly import CCryptoBoringSSLShims
-@_implementationOnly import CryptoBoringWrapper
+import CryptoBoringWrapper
 import Foundation
 
 extension BoringSSLAEAD {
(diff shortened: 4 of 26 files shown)


Lukasa:
cc @0xTim who may want to page some others to test this change

0xTim:
@ptoffy want to test this with JWTKit?

ptoffy:
Well once https://github.com/vapor/jwt-kit/pull/179 lands (which is just waiting on a Crypto release) we won't need any of this anymore. However we can try it out as an assurance for this PR if needed

simonjbeaumont:
> Well once [vapor/jwt-kit#179](https://github.com/vapor/jwt-kit/pull/179) lands (which is just waiting on a Crypto release) we won't need any of this anymore. However we can try it out as an assurance for this PR if needed

@ptoffy if it's not too much hassle, we'd really appreciate it. JWTKit was one place we showed up an issue last time but it wasn't the only one. I _think_ my test of penny-bot was exercising the same thing but I'd love to get as much assurance as we can on this one 🙏

simonjbeaumont:
@ptoffy Just wondering if you had a chance to give this a whirl yet?

ptoffy:
@simonjbeaumont hey yeah I'll look into this today. However I realised I got confused with the crypto PR in which I worked on this and thought we were using this stuff in JWTKit, but we were actually using the BigInt package and Crypto's Arbitrary precision stuff is actually not even public 😅 so I think best I can do is just try to compile the project with this branch

ptoffy:
Okay so I've successfully built JWTKit and ran all tests with your Crypto branch both locally (with a 6.0 compiler) and on a `6.0-jammy` dev container. Is there anything else I can try?

simonjbeaumont:
@ptoffy thanks for confirming! 🙏 

@diegotl @GaultieriOS just wanted to also give you folks a heads up on this, since you both reported issues with your projects from the 3.5.0 release. Would you mind taking a minute to test your project with this branch?

@MahdiBM I know you also had issues with penny-bot. I did test that (see above) but you may wish to do a more thorough pass?

Thanks in advance!

MahdiBM:
@simonjbeaumont thanks for the tests you've already done. I think that should be enough. I don't want to block the PR given that I'm in a place with poor internet connection 🙂.

simonjbeaumont:
@Lukasa I've rebased this on main to resolve the conflicts following the recent PSS key support. PTAL.

Lukasa (approved):

PR #267, opened 2024-10-02
Add ML-DSA post-quantum signatures to `_CryptoExtras`

fpseverino:
Add support for ML-DSA post-quantum digital signatures inside `_CryptoExtras`.

### Checklist

### Motivation:

One of these algorithms is ML-DSA (AKA Dilithium), a module lattice-based signature scheme standardized by NIST in [FIPS 204](https://csrc.nist.gov/pubs/fips/204/final), that is available inside BoringSSL.

By including ML-DSA inside Swift Crypto, we can get closer to normalizing quantum secure algorithms and start implementing them into our apps and libraries to make them quantum-proof.

### Modifications:

Added a `MLDSA65` enum inside the `_CryptoExtras` module with corresponding `PrivateKey`, `PublicKey` and `Signature` structs that use BoringSSL methods to produce and verify ML-DSA-65 digital signatures, with the code style of other signature schemes in the library.

Added tests that cover use cases of the ML-DSA scheme, including [test vectors taken from the BoringSSL repo](https://boringssl.googlesource.com/boringssl/+/refs/heads/master/crypto/mldsa/mldsa_nist_keygen_tests.txt) (extracted from a `.txt` file and encoded in JSON).

### Result:

ML-DSA-65 digital signatures can be created and verified with Swift Crypto.


Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,319 @@
...
+@_implementationOnly import CCryptoBoringSSLShims
+
+/// Types associated with the ML-DSA-65 algorithm
+@_documentation(visibility: public)
+public enum MLDSA {}
+
+extension MLDSA {
+    /// A ML-DSA-65 private key.
+    public struct PrivateKey: Sendable {
+        fileprivate let pointer: UnsafeMutablePointer<MLDSA65_private_key>
```
Lukasa, line 28: This is unsound. It's never possible for a copyable `struct` to own an unsafe pointer safely. This needs to have a backing class to own the pointer.

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,319 @@
...
+        /// - Parameter seed: The seed to use to generate the private key.
+        /// 
+        /// - Throws: `CryptoKitError.incorrectKeySize` if the seed is not at least 32 bytes long or ``CryptoMLDSAError/keyGenerationFailure`` if the key could not be generated.
+        public init(from seed: some DataProtocol) throws {
+            guard seed.count >= MLDSA.seedSizeInBytes else {
+                throw CryptoKitError.incorrectKeySize
+            }
+
+            let seedPtr = UnsafeMutablePointer<UInt8>.allocate(capacity: MLDSA.seedSizeInBytes)
+            seedPtr.initialize(from: seed.regions.flatMap { $0 }, count: MLDSA.seedSizeInBytes)
```
Lukasa, line 66: There's a [`copyBytes(to:)`](https://developer.apple.com/documentation/foundation/dataprotocol/3139158-copybytes) function we can use for this.

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,319 @@
...
+                        contextPointer,
+                        contextLength
+                    )
+                }
+                return rc == 1
+            }
+        }
+
+        /// The size of the public key in bytes.
+        public static let bytesCount = 1952
```
Lukasa, line 272: We should avoid making this public.

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,319 @@
...
+                throw CryptoMLDSAError.keyGenerationFailure
+            }
+        }
+
+        /// Initialize a ML-DSA-65 private key from a DER representation.
+        /// 
+        /// - Parameter derRepresentation: The DER representation of the private key.
+        /// 
+        /// - Throws: `CryptoKitError.incorrectKeySize` if the DER representation is not the correct size or ``CryptoMLDSAError/keyGenerationFailure`` if the key could not be generated.
+        public init(derRepresentation: some DataProtocol) throws {
```
Lukasa, line 85: This isn't parsing the DER representation, this is closer to the `rawRepresentation`.

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,319 @@
...
+
+            guard CCryptoBoringSSL_MLDSA65_parse_private_key(self.pointer, cbsPointer) == 1 else {
+                throw CryptoMLDSAError.keyGenerationFailure
+            }
+        }
+
+        /// Initialize a ML-DSA-65 private key from a PEM representation.
+        /// 
+        /// - Parameter pemRepresentation: The PEM representation of the private key.
+        public init(pemRepresentation: String) throws {
```
Lukasa, line 106: Similarly, there's no PEM here.

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,319 @@
...
+            let document = try ASN1.PEMDocument(pemString: pemRepresentation)
+            self = try .init(derRepresentation: document.derBytes)
+        }
+
+        /// The public key associated with this private key.
+        public var publicKey: PublicKey {
+            get throws {
```
Lukasa, line 113: I don't think there's any reason for this to throw: we should always be able to produce the public counterpart for a private key.

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,319 @@
...
+
+        /// The size of the private key in bytes.
+        public static let bytesCount = 4032
+    }
+}
+
+extension MLDSA {
+    /// A ML-DSA-65 public key.
+    public struct PublicKey: Sendable {
+        private let pointer: UnsafeMutablePointer<MLDSA65_public_key>
```
Lukasa, line 158: Same note on this pointer.

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,319 @@
...
+                    throw CryptoMLDSAError.signatureGenerationFailure
+                }
+
+                length = Signature.bytesCount
+            }
+            return Signature(signatureBytes: output)
+        }
+
+        /// The size of the private key in bytes.
+        public static let bytesCount = 4032
```
Lukasa, line 151: Let's not offer this API publicly.

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,319 @@
...
+        /// 
+        /// - Parameters: 
+        ///   - data: The message to sign.
+        ///   - context: The context to use for the signature.
+        /// 
+        /// - Returns: The signature of the message.
+        public func signature(for data: some DataProtocol, context: [UInt8]? = nil) throws -> Signature {
+            let output = try Array<UInt8>(unsafeUninitializedCapacity: Signature.bytesCount) { bufferPtr, length in
+                let result = data.regions.first!.withUnsafeBytes { dataPtr in
+                    let contextPointer = context?.withUnsafeBufferPointer { $0.baseAddress }
```
Lukasa, line 128: This is unsound: you must not escape this pointer from the closure.

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,319 @@
...
+            }
+        }
+
+        /// Initialize a ML-DSA-65 public key from a seed.
+        /// 
+        /// The seed must be at least 32 bytes long.
+        /// Any additional bytes in the seed are ignored.
+        /// 
+        /// - Parameter seed: The seed to use to generate the public key.
+        public init(from seed: some DataProtocol) throws {
```
Lukasa, line 177: I don't think this API makes sense: users should construct the private key, not the public one.

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,319 @@
...
+            try self.init(privateKey: PrivateKey(from: seed))
+        }
+
+        /// Initialize a ML-DSA-65 public key from a DER representation.
+        /// 
+        /// - Parameter derRepresentation: The DER representation of the public key.
+        /// 
+        /// - Throws: `CryptoKitError.incorrectKeySize` if the DER representation is not the correct size or ``CryptoMLDSAError/keyGenerationFailure`` if the key could not be generated.
+        public init(derRepresentation: some DataProtocol) throws {
```
Lukasa, line 186: This isn't parsing the DER format, it's more like the raw format.

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,319 @@
...
+            guard derRepresentation.count == MLDSA.PublicKey.bytesCount else {
+                throw CryptoKitError.incorrectKeySize
+            }
+
+            self.pointer = UnsafeMutablePointer<MLDSA65_public_key>.allocate(capacity: 1)
+
+            let cbsPointer = UnsafeMutablePointer<CBS>.allocate(capacity: 1)
+            defer { cbsPointer.deallocate() }
+            derRepresentation.regions.flatMap { $0 }.withUnsafeBufferPointer { buffer in
+                cbsPointer.pointee = CBS(data: buffer.baseAddress, len: buffer.count)
```
Lukasa, line 196: You can't do this: this propagates the pointer for the bytes out of the closure, which is forbidden. We need to use the CBS within this closure.

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,319 @@
...
+            guard derRepresentation.count == MLDSA.PrivateKey.bytesCount else {
+                throw CryptoKitError.incorrectKeySize
+            }
+
+            self.pointer = UnsafeMutablePointer<MLDSA65_private_key>.allocate(capacity: 1)
+
+            let cbsPointer = UnsafeMutablePointer<CBS>.allocate(capacity: 1)
+            defer { cbsPointer.deallocate() }
+            derRepresentation.regions.flatMap { $0 }.withUnsafeBufferPointer { buffer in
+                cbsPointer.pointee = CBS(data: buffer.baseAddress, len: buffer.count)
```
Lukasa, line 95: This is unsound: it escapes `buffer` from the closure, which is forbidden. You must use the CBS entirely within this closure.

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,319 @@
...
+
+            guard CCryptoBoringSSL_MLDSA65_parse_public_key(self.pointer, cbsPointer) == 1 else {
+                throw CryptoMLDSAError.keyGenerationFailure
+            }
+        }
+
+        /// Initialize a ML-DSA-65 public key from a PEM representation.
+        /// 
+        /// - Parameter pemRepresentation: The PEM representation of the public key.
+        public init(pemRepresentation: String) throws {
```
Lukasa, line 207: There is no PEM representation here.

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,319 @@
...
+            let document = try ASN1.PEMDocument(pemString: pemRepresentation)
+            self = try .init(derRepresentation: document.derBytes)
+        }
+
+        /// The DER representation of the public key.
+        public var derRepresentation: Data {
```
Lukasa, line 213: Same note here, there's no DER representation.

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,319 @@
...
+            get throws {
+                let cbbPointer = UnsafeMutablePointer<CBB>.allocate(capacity: 1)
```
Lukasa, line 215: We don't ever need to heap allocate the `CBB` structures, we can stack allocate them: `var cbb = CBB()`

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,319 @@
...
+                defer { cbbPointer.deallocate() }
+
+                guard CCryptoBoringSSL_CBB_init(cbbPointer, MLDSA.PublicKey.bytesCount) == 1 else {
```
Lukasa, line 218: `CBB_init` can only return 0 on allocation failure, which we define as impossible. So we can force-unwrap this with a comment to that effect.

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,319 @@
...
+                guard let data = CCryptoBoringSSL_CBB_data(cbbPointer) else {
+                    CCryptoBoringSSL_CBB_cleanup(cbbPointer)
+                    throw CryptoMLDSAError.keyGenerationFailure
+                }
+                return Data(bytes: data, count: CCryptoBoringSSL_CBB_len(cbbPointer))
+            }
+        }
+
+        /// The PEM representation of the public key.
+        public var pemRepresentation: String {
```
Lukasa, line 237: Same note here, there's no PEM representation.

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,319 @@
...
+        /// - Parameters:
+        ///   - signature: The signature to verify.
+        ///   - data: The message to verify the signature against.
+        ///   - context: The context to use for the signature verification.
+        /// 
+        /// - Returns: `true` if the signature is valid, `false` otherwise.
+        public func isValidSignature(_ signature: Signature, for data: some DataProtocol, context: [UInt8]? = nil) -> Bool {
+            return signature.withUnsafeBytes { signaturePtr in
+                let rc: CInt = data.regions.first!.withUnsafeBytes { dataPtr in
+                    let contextPointer = context?.withUnsafeBufferPointer { $0.baseAddress }
```
Lukasa, line 254: Same note here, you can't escape this pointer.

Sources/_CryptoExtras/Util/Error.swift
```diff
@@ -15,3 +15,11 @@
 public enum _CryptoRSAError: Error {
     case invalidPEMDocument
 }
+
+/// Errors that can be thrown when working with ``MLDSA``.
+public enum CryptoMLDSAError: Error {
```
Lukasa, line 20: Let's not add a new error enum here, can we use our existing errors?

fpseverino: Can we use `CryptoKitError.underlyingCoreCryptoError` when BoringSSL fails to generate a key or sign?

fpseverino: Nevermind, I found `internalBoringSSLError()` 👍🏻

fpseverino:
Hi @Lukasa, thank you very much for the quick feedback and directions.

I think I fixed most of your requested changes, but I have some problems with the DER and PEM representations.
Unlike RSA, where there are clear BoringSSL methods for generating and parsing them, I can't find any for ML-DSA. Maybe there are some generic ones? Or a pure Swift implementation is possible/required?

Please let me know if the other changes I made are valid and how I could fix the PEM/DER issue.
Thanks again for the help!

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,428 @@
...
+        /// - Returns: The signature of the message.
+        public func signature(for data: some DataProtocol, context: [UInt8]? = nil) throws -> Signature {
+            try self.backing.signature(for: data, context: context)
+        }
+
+        /// The size of the private key in bytes.
+        static let bytesCount = Backing.bytesCount
+
+        fileprivate final class Backing {
+            let pointer: UnsafeMutablePointer<MLDSA65_private_key>
```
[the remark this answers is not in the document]
fpseverino: So should I add a method to `PrivateKey.Backing` like this one?
```swift
func withUnsafePointer<T>(_ body: (UnsafePointer<MLDSA65_private_key>) throws -> T) rethrows -> T {
    try body(self.pointer)
}
```

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,428 @@
...
+            /// 
+            /// - Parameters: 
+            ///   - data: The message to sign.
+            ///   - context: The context to use for the signature.
+            /// 
+            /// - Returns: The signature of the message.
+            func signature(for data: some DataProtocol, context: [UInt8]? = nil) throws -> Signature {
+                let output = try Array<UInt8>(unsafeUninitializedCapacity: Signature.bytesCount) { bufferPtr, length in
+                    let result = data.regions.first!.withUnsafeBytes { dataPtr in
+                        if let context {
```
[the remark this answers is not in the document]
fpseverino: I'm having some trouble converting the optional `DataProtocol` to an optional `ContiguousBytes`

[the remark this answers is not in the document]
fpseverino: Would this be ok?
```swift
let contextOptional: ContiguousBytes? = if let context { Data(context) } else { nil }
return contextOptional.withUnsafeBytes { contextPtr in
    CCryptoBoringSSL_MLDSA65_verify(
        self.pointer,
        signaturePtr.baseAddress,
        signaturePtr.count,
        dataPtr.baseAddress,
        dataPtr.count,
        contextPtr.baseAddress,
        contextPtr.count
    )
}
```

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,352 @@
...
+
+@_implementationOnly import CCryptoBoringSSL
+@_implementationOnly import CCryptoBoringSSLShims
+
+/// A lattice-based digital signature algorithm that provides security against quantum computing attacks.
+public enum MLDSA {}
+
+extension MLDSA {
+    /// A ML-DSA-65 private key.
+    public struct PrivateKey: Sendable {
```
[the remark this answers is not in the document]
fpseverino: BoringSSL doesn't offer a public API for the serialization of the private key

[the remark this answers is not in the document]
Lukasa: That's extremely weird. Ah well, we should just write it ourselves. It's not a terribly complex format.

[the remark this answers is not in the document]
Lukasa: Ok, I chatted with the BoringSSL folks briefly. It turns out the jury is out on exactly what the serialization format will be for these keys. As a result, let's follow their lead and for now only offer APIs to and from _seed_. We can add other formats when it becomes clear which ones are actually needed.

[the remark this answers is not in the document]
fpseverino: So should I remove all the `rawRepresentation` inits and properties on private and public keys?

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,352 @@
...
+            /// 
+            /// - Throws: `CryptoKitError.incorrectKeySize` if the raw representation is not the correct size.
+            init(rawRepresentation: some DataProtocol) throws {
+                guard rawRepresentation.count == MLDSA.PrivateKey.Backing.bytesCount else {
+                    throw CryptoKitError.incorrectKeySize
+                }
+
+                self.pointer = UnsafeMutablePointer<MLDSA65_private_key>.allocate(capacity: 1)
+
+                try rawRepresentation.regions.flatMap { $0 }.withUnsafeBufferPointer { buffer in
```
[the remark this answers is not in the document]
fpseverino: When I rewrite it like
```swift
let bytes: ContiguousBytes = rawRepresentation.regions.count == 1 ? rawRepresentation.regions.first! : Array(rawRepresentation)
try bytes.withUnsafeBytes { buffer in
    var cbs = CBS(data: buffer.baseAddress, len: buffer.count)
    guard CCryptoBoringSSL_MLDSA65_parse_private_key(self.pointer, &cbs) == 1 else {
        throw CryptoKitError.internalBoringSSLError()
    }
}
```
an error on the CBS init appears, on `buffer.baseAddress`: `Cannot convert value of type 'UnsafeRawPointer?' to expected argument type 'UnsafePointer<UInt8>?' because initializer 'init(data:len:)' was not imported from C header`.

Changing it to `CBS(data: buffer.baseAddress?.assumingMemoryBound(to: UInt8.self), len: buffer.count)` seems to work, is it fine?

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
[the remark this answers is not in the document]
Lukasa: Oh sorry, no, only on private keys. For public keys, the raw representation is what we need to use for now.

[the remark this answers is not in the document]
Lukasa: As there's no "public key from seed" construction.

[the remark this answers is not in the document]
fpseverino: Ok, so on public keys we need an initializer from raw and a property that returns the raw, whereas on the private key I have to add a property that returns the seed, right?

We could save the seed in the `PrivateKey` struct, as it is returned from the BoringSSL init.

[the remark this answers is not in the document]
Lukasa: Yep, that's exactly right.

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,321 @@
...
+            private let seedPointer: UnsafeMutablePointer<UInt8>
+
+            func withUnsafePointer<T>(_ body: (UnsafePointer<MLDSA65_private_key>) throws -> T) rethrows -> T {
+                try body(self.pointer)
+            }
+
+            /// Initialize a ML-DSA-65 private key from a random seed.
+            init() throws {
+                self.pointer = UnsafeMutablePointer<MLDSA65_private_key>.allocate(capacity: 1)
+                self.seedPointer = UnsafeMutablePointer<UInt8>.allocate(capacity: MLDSA.seedSizeInBytes)
```
Lukasa, line 81: I think we can make this a touch more efficient. We can store both of these fields in their useful formats. For the private key, we can just have a `var key: MLDSA65_private_key`, and then derive pointers as needed.

For the seed, we can use a `withUnsafeTemporaryAllocation` and then directly create a `Data` from it. `Data` is the format we need elsewhere anyway.

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,321 @@
...
+            /// Initialize a ML-DSA-65 private key from a seed.
+            /// 
+            /// The seed must be at least 32 bytes long.
+            /// Any additional bytes in the seed are ignored.
+            /// 
+            /// - Parameter seed: The seed to use to generate the private key.
+            /// 
+            /// - Throws: `CryptoKitError.incorrectKeySize` if the seed is not at least 32 bytes long.
+            init(seed: some DataProtocol) throws {
+                guard seed.count >= MLDSA.seedSizeInBytes else {
```
Lukasa, line 103: I can't see any reason to allow the seed to be too long.

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,321 @@
...
+        /// - Returns: `true` if the signature is valid, `false` otherwise.
+        public func isValidSignature<D: DataProtocol>(_ signature: Signature, for data: D, context: D? = nil) -> Bool {
+            self.backing.isValidSignature(signature, for: data, context: context)
+        }
+
+        /// The size of the public key in bytes.
+        static let bytesCount = Backing.bytesCount
+
+        fileprivate final class Backing {
+            private let pointer: UnsafeMutablePointer<MLDSA65_public_key>
```
Lukasa, line 209: Same note here, we can store the `MLDSA64_public_key` directly inline here.

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,321 @@
...
+                        }
+                    }
+                }
+            }
+
+            /// The raw binary representation of the public key.
+            var rawRepresentation: Data {
+                var cbb = CBB()
+                // The following BoringSSL functions can only fail on allocation failure, which we define as impossible.
+                CCryptoBoringSSL_CBB_init(&cbb, MLDSA.PublicKey.Backing.bytesCount)
```
Lukasa, line 245: This currently leaks, we need a `CBB_cleanup`.

Lukasa (commented):
Nice, we're getting a lot closer here!

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
fpseverino: This means that we don't need a `Backing` class?

Lukasa: Technically it does, but we should keep it to avoid us abusing the power of `@_implementationOnly`.

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -16,9 +16,8 @@ import Crypto
 import Foundation
 
 @_implementationOnly import CCryptoBoringSSL
-@_implementationOnly import CCryptoBoringSSLShims
 
-/// A lattice-based digital signature algorithm that provides security against quantum computing attacks.
+/// A module lattice-based digital signature algorithm that provides security against quantum computing attacks.
```
Lukasa, line 20: Do we want `module` here?

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -68,25 +67,26 @@ extension MLDSA {
...
-                self.seedPointer = UnsafeMutablePointer<UInt8>.allocate(capacity: MLDSA.seedSizeInBytes)
-
-                try withUnsafeTemporaryAllocation(of: UInt8.self, capacity: MLDSA.PublicKey.Backing.bytesCount) { publicKeyPtr in
-                    guard CCryptoBoringSSL_MLDSA65_generate_key(
-                        publicKeyPtr.baseAddress,
-                        self.seedPointer,
-                        self.pointer
-                    ) == 1 else {
-                        throw CryptoKitError.internalBoringSSLError()
+                (self.key, self.seed) = try withUnsafeTemporaryAllocation(of: MLDSA65_private_key.self, capacity: 1) { privateKeyPtr in
```
Lukasa, line 75: I don't think we need an `unsafeTemporaryAllocation`. We can just zero-initialize the value `self.key = .init()` and then take a pointer to it.

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -101,24 +101,18 @@ extension MLDSA {
                     throw CryptoKitError.incorrectKeySize
                 }
 
-                self.pointer = UnsafeMutablePointer<MLDSA65_private_key>.allocate(capacity: 1)
-                self.seedPointer = UnsafeMutablePointer<UInt8>.allocate(capacity: MLDSA.seedSizeInBytes)
-
-                let seedBytes = Array(seed.prefix(MLDSA.seedSizeInBytes))
-                seedPointer.initialize(from: seedBytes, count: MLDSA.seedSizeInBytes)
+                self.key = try withUnsafeTemporaryAllocation(of: MLDSA65_private_key.self, capacity: 1) { privateKeyPtr in
```
Lukasa, line 104: Same note about zero-initialization here.

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -203,12 +197,12 @@ extension MLDSA {
...
 
         fileprivate final class Backing {
-            private let pointer: UnsafeMutablePointer<MLDSA65_public_key>
+            var key: MLDSA65_public_key
 
             init(privateKeyBacking: PrivateKey.Backing) {
-                self.pointer = UnsafeMutablePointer<MLDSA65_public_key>.allocate(capacity: 1)
-                let _ = privateKeyBacking.withUnsafePointer { privateKeyPtr in
-                    CCryptoBoringSSL_MLDSA65_public_from_private(self.pointer, privateKeyPtr)
+                self.key = withUnsafeTemporaryAllocation(of: MLDSA65_public_key.self, capacity: 1) { publicKeyPtr in
```
Lukasa, line 203: Same note about zero-initialization here.

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
fpseverino: I added it just to complete the ML-DSA acronym, maybe I could add an hyphen between it and `lattice` to make it clearer

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
fpseverino: Got it. I also tried to remove the seed's `unsafeTemporaryAllocation` by passing to the BoringSSL function an empty `[UInt8]` var and then initializing the `Data` object from it, but I encountered some seg faults in testing

Sources/_CryptoExtras/MLDSA/MLDSA_boring.swift
```diff
@@ -0,0 +1,305 @@
...
+                guard seed.count == MLDSA.seedSizeInBytes else {
+                    throw CryptoKitError.incorrectKeySize
+                }
+
+                self.key = .init()
+                self.seed = Data(seed)
+
+                guard CCryptoBoringSSL_MLDSA65_private_key_from_seed(
+                    &self.key,
+                    Array(seed.prefix(MLDSA.seedSizeInBytes)),
```
Lukasa, line 105: Let's not create an array here, nor should we use `prefix`. We know the seed size is right, and we can use `self.seed.withUnsafeBytes` to get the interior pointer there.

Lukasa (commented):
Cool, looking really good here. One note in the diff.

Lukasa (approved):
Great, I think this is in good shape. Let's get it merged!

Lukasa:
Just a few CI jobs to fix up. Missing license headers, and a need to run the cmake script in the `scripts` directory.

fpseverino:
The license headers are there, but I think the CI wants a `CONTRIBUTORS.txt`, whereas in this repo there is a `CONTRIBUTORS.md`

fpseverino:
Regarding the CMake script, if I run `sh scripts/update_cmakelists.sh` it removes all the libraries string from the `CMakeLists.txt` files. I suppose that's not the intended outcome.

Here's the output:
<details>

```
Finding source files (*.c *.swift) under /Users/francescopaoloseverino/Documents/GitHub/swift-crypto/Sources/CCryptoBoringSSL
scripts/update_cmakelists.sh: line 58: gfind: command not found

Updated /Users/francescopaoloseverino/Documents/GitHub/swift-crypto/Sources/CCryptoBoringSSL/CMakeLists.txt
Finding source files (*.c *.swift) under /Users/francescopaoloseverino/Documents/GitHub/swift-crypto/Sources/CCryptoBoringSSLShims
scripts/update_cmakelists.sh: line 58: gfind: command not found

Updated /Users/francescopaoloseverino/Documents/GitHub/swift-crypto/Sources/CCryptoBoringSSLShims/CMakeLists.txt
Finding source files (*.c *.swift) under /Users/francescopaoloseverino/Documents/GitHub/swift-crypto/Sources/CryptoBoringWrapper
scripts/update_cmakelists.sh: line 58: gfind: command not found

Updated /Users/francescopaoloseverino/Documents/GitHub/swift-crypto/Sources/CryptoBoringWrapper/CMakeLists.txt
Finding source files (*.c *.swift) under /Users/francescopaoloseverino/Documents/GitHub/swift-crypto/Sources/Crypto
scripts/update_cmakelists.sh: line 58: gfind: command not found

Updated /Users/francescopaoloseverino/Documents/GitHub/swift-crypto/Sources/Crypto/CMakeLists.txt
Finding source files (*.c *.swift) under /Users/francescopaoloseverino/Documents/GitHub/swift-crypto/Sources/_CryptoExtras
Excluding source paths (*/AES/*.swift) under /Users/francescopaoloseverino/Documents/GitHub/swift-crypto/Sources/_CryptoExtras
scripts/update_cmakelists.sh: line 58: gfind: command not found

Updated /Users/francescopaoloseverino/Documents/GitHub/swift-crypto/Sources/_CryptoExtras/CMakeLists.txt
Finding assembly files (.S) under /Users/francescopaoloseverino/Documents/GitHub/swift-crypto/Sources/CCryptoBoringSSL
scripts/update_cmakelists.sh: line 72: gfind: command not found
scripts/update_cmakelists.sh: line 73: gfind: command not found
scripts/update_cmakelists.sh: line 74: gfind: command not found
scripts/update_cmakelists.sh: line 75: gfind: command not found




Updated /Users/francescopaoloseverino/Documents/GitHub/swift-crypto/Sources/CCryptoBoringSSL/CMakeLists.txt
```

</details>

P.S. I do have the `find` command installed on my computer

Lukasa:
We expect GNU find. You can get it by running `brew install findutils`

fpseverino:
@Lukasa all done, now the jobs should pass

fpseverino:
I saw that the BoringSSL team added another parameter set to ML-KEM. Considering that they could do this for ML-DSA too, in order not to break the API should we put the current implementation inside a namespace that identifies the parameter set? If so I'm open to suggestions for the name, since the following (which is the most immediate one) won't work
```swift
enum MLDSA {
    enum 65 {}
    
    // Possible future ones
    enum 44 {}
    enum 87 {}
}
```

P.S. Each parameter set uses completely different C++ functions and structures

fpseverino:
Hi @Lukasa,
Thanks again for reviewing and approving this PR!
Just checking if it’s possible to get this merged when you have a moment.
Once this is merged, it would be great to start looking at reviewing #278 and #314.
Let me know if there’s anything else I need to do 🙂

Sources/_CryptoExtras/MLDSA/MLDSA65_boring.swift
```diff
@@ -0,0 +1,320 @@
...
+
+            /// The size of the public key in bytes.
+            static let bytesCount = 1952
+        }
+    }
+}
+
+extension MLDSA65 {
+    /// A ML-DSA-65 signature.
+    public struct Signature: Sendable, ContiguousBytes {
```
Lukasa, line 285: I don't think this type is pulling its weight, honestly. Right now there's only a "raw" ML-DSA signature format, and so we should treat this more like we treat Curve25519, where the output of the signing operation is just `Data`.

fpseverino:
There's an annoying thing in the `signature` and `isValidSignature` methods that I don't know how to solve:

I wanted to give a separate generic type to the `context` argument, like `Curve25519` (and now also `MLDSA65`) do for the `data` and `signature` arguments, but since it is `nil` by the default the compiler can't infer it.
The solution in place right now is to use the same generic type as `data`.

Sources/_CryptoExtras/MLDSA/MLDSA65_boring.swift
```diff
@@ -0,0 +1,300 @@
...
+        ///
+        /// - Parameter seed: The seed to use to generate the private key.
+        ///
+        /// - Throws: `CryptoKitError.incorrectKeySize` if the seed is not 32 bytes long.
+        public init(seed: some DataProtocol) throws {
+            self.backing = try Backing(seed: seed)
+        }
+
+        /// The seed from which this private key was generated.
+        public var seed: Data {
```
Lukasa, line 42: Let's call this `seedRepresentation`.

Lukasa (commented):
Thanks for this, and sorry for the delay in review! Generally I'm really happy with how this looks. I've left a few notes in the diff.

Sources/_CryptoExtras/MLDSA/MLDSA65_boring.swift
```diff
@@ -0,0 +1,300 @@
...
+        /// - Returns: The signature of the message.
+        public func signature<D: DataProtocol>(for data: D, context: D? = nil) throws -> Data {
+            try self.backing.signature(for: data, context: context)
+        }
+
+        /// The size of the private key in bytes.
+        static let byteCount = Backing.byteCount
+
+        fileprivate final class Backing {
+            var key: MLDSA65_private_key
```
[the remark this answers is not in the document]
fpseverino: I have to make this `fileprivate` because it's used in the public key backing

Lukasa (commented):
OK, two very minor nits and then I think we're good to go. I'll also kick the tests off here so we have that information.

Lukasa (approved):
Ok, this looks really nice. Thanks for your hard work here @fpseverino! :sparkles:

fpseverino:
Thank you @Lukasa for all the help and guidance you have given me!

PR #268, opened 2024-10-02
Support loading RSAPSS public keys with parameters

gautierdelorme:
Support loading RSA PSS public keys with parameters.

### Checklist

### Motivation:

Trying to load those keys currently fails.

### Modifications:

Parameters are stripped and the key is treated as a regular public key.

### Result:

`_RSA.Signing.PublicKey()` works with RSA PSS keys with parameters.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -85,7 +85,7 @@ let package = Package(
             MANGLE_END */
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-asn1.git", .upToNextMajor(from: "1.2.0"))
+        .package(url: "https://github.com/apple/swift-asn1.git", from: "1.2.0")
     ],
     targets: [
         .target(
--- a/Package.swift
+++ b/Package.swift
@@ -84,7 +84,9 @@ let package = Package(
             .library(name: "CCryptoBoringSSL", type: .static, targets: ["CCryptoBoringSSL"]),
             MANGLE_END */
     ],
-    dependencies: [],
+    dependencies: [
+        .package(url: "https://github.com/apple/swift-asn1.git", .upToNextMajor(from: "1.2.0"))
+    ],
     targets: [
         .target(
             name: "CCryptoBoringSSL",
@@ -141,7 +143,8 @@ let package = Package(
                 "CCryptoBoringSSL",
                 "CCryptoBoringSSLShims",
                 "CryptoBoringWrapper",
-                "Crypto"
+                "Crypto",
+                .product(name: "SwiftASN1", package: "swift-asn1")
             ],
             exclude: privacyManifestExclude + [
                 "CMakeLists.txt",
--- a/Sources/_CryptoExtras/RSA/RSA.swift
+++ b/Sources/_CryptoExtras/RSA/RSA.swift
@@ -13,6 +13,7 @@
 //===----------------------------------------------------------------------===//
 import Foundation
 import Crypto
+import SwiftASN1
 
 #if CRYPTO_IN_SWIFTPM && !CRYPTO_IN_SWIFTPM_FORCE_BUILD_API
 fileprivate typealias BackingPublicKey = SecurityRSAPublicKey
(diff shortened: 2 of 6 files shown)


Package.swift
```diff
@@ -84,7 +84,9 @@ let package = Package(
             .library(name: "CCryptoBoringSSL", type: .static, targets: ["CCryptoBoringSSL"]),
             MANGLE_END */
     ],
-    dependencies: [],
+    dependencies: [
+        .package(url: "https://github.com/apple/swift-asn1.git", .upToNextMajor(from: "1.2.0"))
```
Lukasa, line 88: Nit:

```suggestion
        .package(url: "https://github.com/apple/swift-asn1.git", from: "1.2.0")
```

Lukasa:

Sources/_CryptoExtras/Util/SubjectPublicKeyInfo.swift
```diff
@@ -0,0 +1,116 @@
+//
+//  SubjectPublicKeyInfo.swift
+//
+//
+//  Created by Gautier Delorme on 24/09/2024.
+//
```
Lukasa, line 6: Can you add the license header that is present in other files and remove this header?

Sources/_CryptoExtras/RSA/RSA.swift
```diff
@@ -95,9 +96,12 @@ extension _RSA.Signing {
         ///
         /// This constructor supports key sizes of 1024 bits or more. Users should validate that key sizes are appropriate
         /// for their use-case.
+        /// Parameters from RSA PSS keys will be stripped.
         /// - Warning: Key sizes less than 2048 are not recommended and should only be used for compatibility reasons.
         public init<Bytes: DataProtocol>(unsafeDERRepresentation derRepresentation: Bytes) throws {
-            self.backing = try BackingPublicKey(derRepresentation: derRepresentation)
+            let sanitizedDer = try SubjectPublicKeyInfo.stripRsaPssParameters(derEncoded: [UInt8](derRepresentation))
```
Lukasa, line 102: Not a request for changes in this PR, but can you file an issue that asks us to move all our RSA key parsing into Swift ASN1? This is a pragmatic change for now, but as we're _already_ parsing the key we should probably just stick with that now.

Sources/_CryptoExtras/Util/SubjectPublicKeyInfo.swift
```diff
@@ -0,0 +1,116 @@
...
+}
+
+extension SubjectPublicKeyInfo {
+    static func stripRsaPssParameters(derEncoded: [UInt8]) throws -> [UInt8] {
+        guard var spki = try? SubjectPublicKeyInfo(derEncoded: derEncoded) else {
+            // If it's not a SPKI then it can't be a PSS key so we just return it.
+            return derEncoded
+        }
+
+        if spki.algorithmIdentifier.algorithm == .AlgorithmIdentifier.rsaPSS {
```
Lukasa, line 107: This doesn't look like it'll compile: I don't think that algorithm identifier is actually defined anywhere in this code.

Lukasa:
Thanks for this PR! I've left a few notes in the diff for cleaning up.

Sources/_CryptoExtras/Util/SubjectPublicKeyInfo.swift
gautierdelorme: It comes from `swift-asn1`: https://github.com/apple/swift-asn1/blob/1a80ac7e4b8831264b2964758732a63265457a58/Sources/SwiftASN1/Basic%20ASN1%20Types/ObjectIdentifier.swift#L259-L263.

Sources/_CryptoExtras/RSA/RSA.swift
gautierdelorme: https://github.com/apple/swift-crypto/issues/269

Sources/_CryptoExtras/Util/SubjectPublicKeyInfo.swift
```diff
@@ -0,0 +1,124 @@
...
```
Lukasa, line 114: Can we turn this into a shortcut early return? If we don't have to make any modifications, we don't need to re-encode the key.

Lukasa (approved):
Cool, this LGTM.

Lukasa:
Thanks very much!

Lukasa (approved):

PR #270, opened 2024-10-04
Update code to 2024 SDK release

Lukasa:
This patch updates Crypto to the 2024 release version of CryptoKit. As with prior years, there are small tweaks to the API surface that cause this to manifest as a semver major. In this instance, the relevant change is the removal of the long-deprecated setters for the hash function block byte counts.

This is another change that meets the technical definition of a semver major, but is practically extraordinarily unlikely to manifest in an actual problem. Nonetheless, we do have to acknowledge the reality that this can break compiling code (e.g. in cases where users have defined protocols that rely on having a setter available, even though they never call through it).

To that end, this pushes Crypto up to 4.0.


Lukasa:
cc @0xTim @ptoffy @gwynne as major downstream customers who may need to be aware of the pending semver major here.

PR #277, opened 2024-10-15
wrapper: Move EC types and random bytes into CryptoBoringWrapper

simonjbeaumont:
### Motivation

The Crypto module contains some EC types used to implement the CryptoKit API on non-Darwin platforms, which we'd like to make available to functionality in _CryptoExtras on all platforms.

### Modifications

In a similar manner to how we handled `ArbitraryPrecisionInteger`, this PR moves the EC types, which were internal to the Crypto module, to the CryptoBoringWrapper module, and makes them package visible.

This patch only exposes the functionality that's actually used by the Crypto module as package-visible, and leaves everything else internal (now internal to the CryptoBoringWrapper module).

As part of the move, the `_boring` suffix has been removed, since that is implied by their presence in this module. This PR also retroactively removes that suffix from the other files that migrated over recently but erroneously retained their suffix.

Finally, we remove some redundant compilation guards on the tests for files that were previously moved into `CryptoBoringWrapper` since they are intended to run on all platforms now.

### Result

This should have no external impact and is purely a refactor.


Lukasa (approved):
Thanks Si.

PR #278, opened 2024-10-15
Add SLH-DSA post-quantum signatures to CryptoExtras

fpseverino:
Add support for SLH-DSA post-quantum digital signatures inside CryptoExtras.

### Checklist

### Motivation:

One of these algorithms is SLH-DSA (AKA SPHINCS+), a stateless hash-based signature scheme standardized by NIST in [FIPS 205](https://csrc.nist.gov/pubs/fips/205/final), that is available inside BoringSSL.

By including SLH-DSA inside Swift Crypto, we can get closer to normalizing quantum secure algorithms and start implementing them into our apps and libraries to make them quantum-proof.

### Modifications:

Added a `SLHDSA` enum inside the `CryptoExtras` module with corresponding `PrivateKey`, `PublicKey` and `Signature` structs that use BoringSSL methods to produce and verify SLH-DSA-SHA2-128s digital signatures, with the code style of other signature schemes in the library.

### Result:

SLH-DSA-SHA2-128s digital signatures can be created and verified with Swift Crypto.


Lukasa:
Thank you for this @fpseverino! I haven't forgotten it: I just want to make sure we get the ML-DSA stuff landed _first_, to maximise my useful review bandwidth. Is that ok?

Lukasa:
Yeah, for DER/PEM until we have a standard with final IANA codepoints it's not a good idea for us to add the code unless someone has a need. Easiest thing to do might be to remove it for now, and open a separate draft PR that adds it back?

Lukasa (commented):
Ok, this is looking great. Can we add the test-vectors from the NIST publication? There seem to be three sets but we'll need to dig through to figure out which ones apply: https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files.

fpseverino:
> > Can we add the test-vectors from the NIST publication?
> 
> For key generation we would need an API to generate keys from seeds, which BoringSSL doesn't offer.

They recently added key generation from seeds in the public API: https://boringssl.googlesource.com/boringssl/+/0934f4a1f929f11cd3a420ac47641aa8fd584552

PR #281, opened 2024-10-16
Rename _CryptoExtras to CryptoExtras

0xTim:
Rename `_CryptoExtras` to `CryptoExtras`

### Checklist

### Motivation:

CryptoExtras is now depended on by a number of libraries and well used and well understood. This PR formalises this with the removal of the underscore to the product to denote it's now a 'real' library

### Modifications:

Renamed the product

### Result:

`import _CryptoExtras` is now `import CryptoExtras`


Package.swift
```diff
@@ -79,7 +79,8 @@ let package = Package(
     ],
     products: [
         .library(name: "Crypto", targets: ["Crypto"]),
-        .library(name: "_CryptoExtras", targets: ["_CryptoExtras"]),
+        .library(name: "_CryptoExtras", targets: ["CryptoExtras"]),
```
0xTim, line 82: One for discussion - the README is set up that this library **won't** exist but I've left it here for discussion if we want to keep it to maintain backwards compatibility or not or release this with the upcoming major version

0xTim:
One thing this PR doesn't currently change is the CMake files - will wait to hear if it should be a breaking or backwards compatible change first

Package.swift
Lukasa: I think retaining the backward compatibility is valuable: while the underscore is present, we've tried to avoid gratuitously breaking things and it'd be nice to do so here.

0xTim: Ok I've updated the README. In terms of getting this to work with CMake do I need an shim for `_CryptoExtras` that points to `CryptoExtras`?

Lukasa:
I'm going to suggest we slow-roll this until we've had the conversation about the best name for this module.

fpseverino (changes requested):
For the CryptoExtras DocC landing page to be rendered, the underscore must also be removed [here](https://github.com/apple/swift-crypto/blob/041199614e658c8e921e7de1b5b589b1fb723a4c/Sources/_CryptoExtras/Docs.docc/index.md?plain=1#L1)

0xTim:
Linking to the forum threads discussion to make CryptoExtras the main library and `CryptoCore` what is currently `Crypto`. Also discussions around package traits

https://forums.swift.org/t/improving-crypto-on-all-platforms/75330/37

fpseverino:
Since it will mainly be used on Apple platforms, to keep it in line with other Apple frameworks (e.g. Core Data, Core Foundation, Core Image), I would rename what is now `Crypto` to `CoreCrypto`.

As has been discussed on the forum, will the new `Crypto` (formerly `_CryptoExtras`) `@_exported import CoreCrypto`?

I do not fully understand the idea of enabling the dependency to BoringSSL with a package trait, how would that work?
But I am definitely in favour of hiding other even more insecure but frequently used algorithms behind a package trait

Lukasa:
`CoreCrypto` is definitely not going to be the name, as that exists (it's an implementation detail of Apple's platforms but its source code is available from developer.apple.com for security audit).

fpseverino:
> `CoreCrypto` is definitely not going to be the name, as that exists (it's an implementation detail of Apple's platforms but its source code is available from developer.apple.com for security audit).

Then to avoid any confusion between `CoreCrypto` and `CryptoCore` we could rename (as has been suggested [here](https://forums.swift.org/t/improving-crypto-on-all-platforms/75330/34)) what is now called `Crypto` as `CryptoEssentials` and `_CryptoExtras` would become `Crypto`.

I believe that, just as with Foundation, people would prefer to use the "Essentials" module by default (which on Apple's platforms would be the best performing), and only `import Crypto` (which would export `CryptoEssentials`) if they need the additional API, which is almost the same thing as today with `Crypto` and `_CryptoExtras`, but this way we would be conforming to Foundation's naming standard and emphasising that the "Extras" APIs are not a "side project".

fpseverino:
I would like to propose [this idea](https://github.com/apple/swift-crypto/pull/281#issuecomment-2855300630) to you again

0xTim:
`ExtendedCrypto` is the best option I've seen if the decision is to keep `Crypto` as is. I'm still in favour of `CryptoEssentials` for my vote but would be happy with `ExtendedCrypto`

fpseverino:
IMHO `ExtendedCrypto` is not much better than `CryptoExtras`, the latter has also become something of a norm in the community (e.g. `swift-nio-extras`, `sql-kit-extras`, etc.).

Also, IMO, whatever the name of the module will be, it would be nicer if it started with `Crypto` and then the eventual suffix

PR #285, opened 2024-10-20
Typealias CryptoKitError

0xTim:
Provides a typealias for `CryptoKitError` to hide the implementation detail of CryptoKit

### Checklist
- [ ] I've followed the code style of the rest of the project
- [ ] I've read the [Contribution Guidelines](CONTRIBUTING.md)

### Motivation:

CryptoKit is an implementation detail on Apple platforms - this makes it clearer for users coming from other platforms

### Modifications:

Added a type alias and updated documentation

### Result:


0xTim:
@Lukasa is there anything more that needs to be done on this?

0xTim:
@Lukasa ok updated the PR to avoid any filename changes and kept the renames to just the API docs (as opposed to changing the code to throw a `CryptoError`.

I think the CryptoExtras code should be updated to remove any references to CryptoKit but that can be done in a separate PR

Sources/Crypto/CryptoError_boring.swift
```diff
@@ -0,0 +1,4 @@
+@available(iOS 13.0, macOS 10.15, watchOS 6.0, tvOS 13.0, *)
```
Lukasa, line 1: Sorry to be a bore here, but you'll need the license header on this file.

Lukasa (commented):
Agreed on the follow-up PR as well.

Sources/Crypto/CryptoError_boring.swift
0xTim: Ah good point, added

Lukasa:
Ah shoot, can you rebuild the CMake pieces? There's a script.

Sources/CCryptoBoringSSL/CMakeLists.txt
```diff
@@ -262,78 +262,16 @@ add_library(CCryptoBoringSSL STATIC
...
-    gen/bcm/ghash-neon-armv8-linux.S
-    gen/bcm/ghashv8-armv8-linux.S
-    gen/bcm/p256-armv8-asm-linux.S
-    gen/bcm/p256_beeu-armv8-asm-linux.S
-    gen/bcm/sha1-armv8-linux.S
-    gen/bcm/sha256-armv8-linux.S
-    gen/bcm/sha512-armv8-linux.S
-    gen/bcm/vpaes-armv8-linux.S
-    gen/crypto/chacha-armv8-linux.S
-    gen/crypto/chacha20_poly1305_armv8-linux.S)
```
Lukasa, line 336: Hmm, this looks incorrect. Oh shoot, I think we did some upstream work on this. Try this one? https://github.com/apple/swift-nio/blob/main/scripts/update-cmake-lists.sh

0xTim: That's the script I copied from when running. Do you get the same result running the script?

fpseverino: I think I had the same problem, do you also get the error `scripts/update_cmakelists.sh: line 58: gfind: command not found` or something similar? In my case the solution was to download GNU find with `brew install findutils`

0xTim: @Lukasa ok I think I've fixed this by pulling in the latest upstream changes

Lukasa (approved):
Great stuff, thanks @0xTim.

PR #289, opened 2024-10-29
Fix Windows CI

ptoffy:
Add a platform check to `scrypt` which replaces `sysconf` with its Windows counterpart to make Windows CI pass


Lukasa (approved):
Thanks @ptoffy!

Lukasa:
And great timing, @FranzBusch just added the windows CI pipeline. Let's see what this looks like.

FranzBusch:
The tests build and run now but we have numerous failures.

Lukasa:
Yeah, we knew that we had CI failures. Paul was only trying to get the build to pass, so we can merge this as-is. WE'll need to fix the Windows CI failures, of course, but at least we now have a CI system that is capable of locking in the win.

PR #292, opened 2024-11-06
extras: Add EC toolbox abstractions and OPRF(P-384, SHA-384) VOPRF API

simonjbeaumont:
## Motivation

We would like to provide support for the P384-SHA384 Verifiable Oblivious Pseudorandom Function (VOPRF) as defined in [RFC 9497: VOPRF Protocol](https://www.rfc-editor.org/rfc/rfc9497.html#name-voprf-protocol).

## Modifications

- Add protocols for some "ECToolbox" abstractions: `Group`, `GroupElement`, `GroupScalar`, `HashToGroup`.
- Add implementations backed by BoringSSL.
- Duplicate some utilities from Crypto into _CryptoExtras: hexadecimal bytes and I2OSP.
- Add VOPRF API, rooted at `enum P384._VOPRF`.

## Results

- New internal abstractions that make it easier to bring up new implementations.
- New API for `OPRF(P-384, SHA-384)` in VOPRF mode.

## Tests

- Test vectors for internal `HashToField` implementation.
- Test vectors for internal VOPRF implementation.
- Tests of the VOPRF public API.


simonjbeaumont:
@Lukasa PTAL when you have time 🙏

Sources/CryptoBoringWrapper/EC/EllipticCurve.swift
```diff
@@ -72,6 +72,14 @@ extension BoringSSLEllipticCurveGroup {
         return try! ArbitraryPrecisionInteger(copying: baseOrder)
     }
 
+    @usableFromInline
+    package var generator: EllipticCurvePoint { get throws {
```
Lukasa, line 76: Might be worth not squishing this in, `get throws {` can be on its own line.

Sources/CryptoBoringWrapper/EC/EllipticCurvePoint.swift
```diff
@@ -12,41 +12,262 @@
...
+        try self.init(copying: other._basePoint, on: group)
+    }
+
+    @usableFromInline
+    package init(multiplying scalar: ArbitraryPrecisionInteger, on group: BoringSSLEllipticCurveGroup) throws {
+        self._basePoint = try scalar.withUnsafeBignumPointer { scalarPtr in
+            try group.withUnsafeGroupPointer { groupPtr in
+                guard
+                    let pointPtr = CCryptoBoringSSL_EC_POINT_new(groupPtr),
+                    CCryptoBoringSSL_EC_POINT_mul(groupPtr, pointPtr, scalarPtr, nil, nil, nil) == 1
```
Lukasa, line 45: Not a new issue, but this leaks the `pointPtr` if the mul fails.

Sources/CryptoBoringWrapper/EC/EllipticCurvePoint.swift
```diff
@@ -12,41 +12,262 @@
...
+        case .p521: throw CryptoBoringWrapperError.invalidParameter // BoringSSL doesn't have a hash_to_curve API for P521.
+        case .none: throw CryptoBoringWrapperError.internalBoringSSLError()
+        }
+
+       self._basePoint = try msg.withUnsafeBytes { msgPtr in
+            try group.withUnsafeGroupPointer { groupPtr in
+                try domainSeparationTag.withUnsafeBytes { dstPtr in
+                    guard
+                        let pointPtr = CCryptoBoringSSL_EC_POINT_new(groupPtr),
+                        hashToCurveFunction(
```
Lukasa, line 176: If this fails, we leak the point.

Sources/CryptoBoringWrapper/EC/EllipticCurvePoint.swift
```diff
@@ -12,41 +12,262 @@
...
+
+    @usableFromInline
+    package func isEqual(to rhs: EllipticCurvePoint, on group: BoringSSLEllipticCurveGroup) -> Bool {
+        self.withPointPointer { selfPtr in
+            group.withUnsafeGroupPointer { groupPtr in
+                rhs.withPointPointer { rhsPtr in
+                    switch CCryptoBoringSSL_EC_POINT_cmp(groupPtr, selfPtr, rhsPtr, nil) {
+                    case 0: return true
+                    case 1: return false
+                    default: fatalError("Unexpected error testing EC point equality")
```
Lukasa, line 199: I don't know that we need to be this strict. `0` is true, everything else is `false`. Happy to have an assertion for the other codes, but probably shouldn't fatal error.

Sources/CryptoBoringWrapper/EC/EllipticCurvePoint.swift
```diff
@@ -12,41 +12,262 @@
...
-            guard let basePoint = CCryptoBoringSSL_EC_POINT_dup(pointer, groupPtr) else {
+            guard let basePoint = CCryptoBoringSSL_EC_POINT_new(groupPtr) else {
                 throw CryptoBoringWrapperError.internalBoringSSLError()
             }
             return basePoint
         }
+
+        guard group.withUnsafeGroupPointer({ groupPtr in
+            bytes.withUnsafeBytes { dataPtr in
+                CCryptoBoringSSL_EC_POINT_oct2point(
```
Lukasa, line 217: If this fails we leak the point.

Sources/CryptoBoringWrapper/EC/EllipticCurvePoint.swift
simonjbeaumont: I think that's not quite right since the docs say that `EC_POINT_cmp` can return an error: 

```
EC_POINT_cmp returns 1 if the points are not equal, 0 if they are, or -1 on error.
```

I think this can happen if you provide a point on a different group for an equality test, for instance.

This function isn't throwing so the best we can do is fatal error, which seems to be in line with what we do elsewhere?

Sources/CryptoBoringWrapper/EC/EllipticCurve.swift
simonjbeaumont: OK, I moved this onto its own line. 

Aside: I have a task to add the swift-format job to this repo soon, which will squash all these inconsistencies away.

Sources/CryptoBoringWrapper/EC/EllipticCurvePoint.swift
simonjbeaumont: I've addressed this by factoring out the `EC_POINT_new` into a designated `init(_pointAtInfinityOn group:)` and made all these `convenience` initialisers.

Sources/CryptoBoringWrapper/EC/EllipticCurvePoint.swift
simonjbeaumont: Resolved in the same manner as https://github.com/apple/swift-crypto/pull/292/files/fda1a2c6df294e87b24b77ad28a7cd5950d74d25#r1850181741.

Sources/CryptoBoringWrapper/EC/EllipticCurvePoint.swift
simonjbeaumont: Resolved in the same manner as https://github.com/apple/swift-crypto/pull/292/files/fda1a2c6df294e87b24b77ad28a7cd5950d74d25#r1850181741.

simonjbeaumont (commented):
Thanks for the feedback @Lukasa. I've addressed your comments. Could you take another look when you get a minute.

Sources/CryptoBoringWrapper/EC/EllipticCurvePoint.swift
simonjbeaumont: After further consideration, given we allow folks to pass in the `group`, we shouldn't even assert, but just return `false` if comparing two points from different groups.

I've updated the code to clear the BoringSSL error and return `false` in this scenario.

simonjbeaumont:
Thanks @Lukasa :) please merge when convenient.

PR #293, opened 2024-11-08
Android build fixes (build with -fPIC)

lhoward:
Build with -fPIC on Android as it is not the default, and is necessary if swift-crypto is to be embedded into a shared library.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -114,6 +114,7 @@ let package = Package(
                  */
                 .define("OPENSSL_NO_THREADS_CORRUPT_MEMORY_AND_LEAK_SECRETS_IF_THREADED", .when(platforms: [Platform.wasi])),
                 .define("OPENSSL_NO_ASM", .when(platforms: [Platform.wasi])),
+                .unsafeFlags(["-fPIC"], .when(platforms: [Platform.android])),
             ]
         ),
         .target(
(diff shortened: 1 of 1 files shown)


Lukasa (changes requested):
We cannot set unsafe flags, unfortunately, as they prevent the library being used as a dependency.

lhoward:
Ah, yes, right you are.

@finagolfin is there a way to ensure the Android SDK is always compiling PIC even when not building shared libraries?

lhoward:
I'm going to close this because it needs to be fixed in the Android SDK.

finagolfin:
> is there a way to ensure the Android SDK is always compiling PIC even when not building shared libraries?

We used to [set it in the destination config until Swift 6](https://github.com/finagolfin/swift-android-sdk/commit/aa00f197c43aeb229233f989e0450f31c62b26f1#diff-2cefc1d2eca9206472cbc8b16427c9a526e6210be5086f0404b942c9a44b1dfdL7), but I thought it was made the default now, so I didn't bother setting it in the new Swift 6 SDK bundle.

If that's not the case, you can always pass in `-Xcc -fPIC` on the command-line, or we can change the default for the Android SDK bundle.

lhoward:
Thanks. Actually I just removed the swift-crypto dependency from my app, was the simplest solution for now :)

PR #297, opened 2024-11-14
add .spi.yml

rnro:
Add .spi.yml for `Crypto` target so that DocC documentation is generated.


glbrntt (approved):

0xTim:
Could we incorporate 283 to add Crypto Extras as well?

PR #299, opened 2024-11-18
Potentially Inconsistent Coordinate Slicing

dimitribouniol:
Replaced potentially unexpected accesses to coordinate slices with consistent ranges no matter if the raw representation is a root or a slice of data.

### Checklist

### Motivation:

As discussed [in slack](https://swift-open-source.slack.com/archives/C04SH3Y3X27/p1731924717212829), `.prefix(upTo:)` and `.suffix(from:)` are problematic when you don't know the origin of `Data`, as a slice of Data can have a non-zero `startIndex`, causing downstream misalignment. For myself, this happened when I tried accessing the `rawRepresentation` of an ECDSA's PublicKey, which was derived from the x963 representation, and thus had a start offset of 1. Although the code performed as expected, this seemed like a mistake waiting to happen as changing how `rawRepresentation` is derived for signatures could cause a similar headache in the future, especially so for anyone reading the code as a reference (it me!).

### Modifications:

Since the use cases here were to split the `Data` in half, I replaces the uses of `.prefix(upTo:)` and `.suffix(from:)` with `.prefix()` and `.suffix()` respectfully.

### Result:

No change to behavior, but more semantically correct code.

### DIFF
--- a/Sources/Crypto/Signatures/ECDSA.swift
+++ b/Sources/Crypto/Signatures/ECDSA.swift
@@ -66,7 +66,7 @@ extension P256.Signing {
             let combined = rawRepresentation
             assert(combined.count % 2 == 0)
             let half = combined.count / 2
-            return (combined.prefix(upTo: half), combined.suffix(from: half))
+            return (combined.prefix(half), combined.suffix(half))
         }
 
         /// Creates a P-256 digital signature from a Distinguished Encoding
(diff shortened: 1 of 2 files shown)


Lukasa (approved):

Lukasa:
Thanks @dimitribouniol! :sparkles:

PR #302, opened 2024-11-26
Enable swift-format in CI

simonjbeaumont:
## Motivation

As part of the Github Actions CI migration we held off enabling the format check and moving to the same formatting rules as Swift NIO, but we'd like to do that now.

## Modifications

- Add a script to generate (and update) the `.swiftformatignore` file since this repo only wants to format a specific set of files matching a pattern.
- Generate and commit the `.swiftformatignore` file.
- Run the fomatter and commit the results.
- Additional commits to satisfy the remaining linter rules that weren't auto-fixed by swift-format.
- Enable the format check in CI.

## Result

- Code that's not shared is formatted with a style consistent with the other repos this team maintain.
- CI is now checking code format again.


.github/workflows/pull_request.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     uses: swiftlang/github-workflows/.github/workflows/soundness.yml@main
     with:
       license_header_check_project_name: "SwiftCrypto"
-      format_check_enabled: false
+      format_check_enabled: true
```
rnro, line 13:
```suggestion
```

It's probably better to just let the default apply in this case.

Lukasa (approved):
Thanks @simonjbeaumont!

rnro (changes requested):
Can we remove the unnecessary parameter?

PR #314, opened 2024-12-14
Add ML-KEM post-quantum key agreement to `_CryptoExtras`

fpseverino:
Add support for ML-KEM post-quantum key encapsulation mechanism inside `_CryptoExtras`.

### Checklist

### Motivation:

One of these algorithms is ML-KEM (AKA Kyber), a module lattice-based key encapsulation mechanism standardized by NIST in [FIPS 203](https://csrc.nist.gov/pubs/fips/203/final), that is available inside BoringSSL.

By including ML-KEM inside Swift Crypto, we can get closer to normalizing quantum secure algorithms and start implementing them into our apps and libraries to make them quantum-proof.

### Modifications:

Added `MLKEM768` and `MLKEM1024` enums (generated with `gyb`) inside the `_CryptoExtras` module with corresponding `PrivateKey` and `PublicKey` structs that use BoringSSL methods to decapsulate and encapsulate ML-KEM-768 and ML-KEM-1024 ciphertexts, with the code style of other KEMs in the library (they also conform to `KEMPrivateKey` and `KEMPublicKey` protocols).

### Result:

ML-KEM key agreements can be handled with Swift Crypto.


Sources/_CryptoExtras/MLKEM/MLKEM_boring.swift.gyb
```diff
@@ -0,0 +1,305 @@
...
+        public static func generate() -> MLKEM${parameter_set}.PrivateKey {
+            .init()
+        }
+
+        /// Initialize a ML-KEM-${parameter_set} private key from a seed.
+        ///
+        /// - Parameter seed: The seed to use to generate the private key.
+        ///
+        /// - Throws: `CryptoKitError.incorrectKeySize` if the seed is not 64 bytes long.
+        public init(seed: some DataProtocol) throws {
```
Lukasa, line 54: Let's call this `seedRepresentation` both here and in the getter.

Sources/_CryptoExtras/MLKEM/MLKEM_boring.swift.gyb
```diff
@@ -0,0 +1,305 @@
...
+        }
+
+        /// Decapsulate a shared secret and create a symmetric key.
+        ///
+        /// - Parameter encapsulated: The encapsulated shared secret.
+        ///
+        /// - Throws: `CryptoKitError.incorrectParameterSize` if the encapsulated shared secret is not 1088 bytes long.
+        ///
+        /// - Returns: The symmetric key.
+        public func decapsulate(_ encapsulated: Data) throws -> SymmetricKey {
```
Lukasa, line 75: Let's make this generic over `DataProtocol`.

fpseverino: I can't if we want to conform to the `KEMPrivateKey` protocol

Lukasa: I think the generic should satisfy the protocol requirement.

Lukasa (approved):
I think this looks good as well, let's see if we can land this. Thanks @fpseverino! :sparkles:

fpseverino:
The only justification for [this error](https://github.com/apple/swift-crypto/actions/runs/14754775256/job/41420863518?pr=314#step:4:53717) (which had never happened) that I can think of is that the first byte of the random seed was already `0xff`, so the derived key is the same. [I modified the test](https://github.com/apple/swift-crypto/pull/314/commits/5fa8835a666281a6f21fdc092fef555a7d1cb63a) to change the first two bytes to `0xff`, so that the probability of this happening again is lower.

PR #315, opened 2024-12-16
FreeBSD support

michael-yuji:
Fix FreeBSD support

### Checklist

### Motivation:
To add FreeBSD support to Swift, we must ensure that swift crypto builds on FreeBSD since it's a dependency of swiftpm and several essential packages.

### Modifications:
- `Package.swift` - Ideally FreeBSD should use `.when(platforms:)` to set the corresponding swift settings and dependencies. However, `Platform.freebsd` is not yet available on swiftpm and therefore use guard the settings behind a `isFreeBSD` boolean flag. This is chosen because it minimize the diff and maintains the maximum readability.
- `cmake/modules/SwiftSupport.cmake` - add `amd64` arch support, maps  to `x86_64`.
- `**/CMakeLists.txt` - Use the Linux assembly files for FreeBSD. Both Linux and FreeBSD uses SysV ABI, and there're no syscalls involved (after all, they're supposed to be "pure") so there are no ABI compatibility problem. After all, all the tests are passing on FreeBSD with this PR.

### Result:
All the tests (via `swift test`) passed on x86_64 FreeBSD 14.1-p6, and passed on aarch64 macOS.

Test log available here: [swift-test-result.txt](https://github.com/user-attachments/files/18147920/swift-test-result.txt)


Lukasa (approved):
Thanks for this! Hopefully we'll get a FreeBSD CI environment to prevent this regressing, but for now this looks totally reasonable. Nice one! :sparkles:

Package.swift
```diff
@@ -27,9 +27,18 @@ import PackageDescription
 // To develop this on Apple platforms, set this to true
 let development = false
 
+// Ideally, we should use `.when(platforms:)` to set `swiftSettings` and
+// `dependencies` like on other platforms. However, `Platform.freebsd` is not
+// yet available, and therefore we guard the settings behind this boolean.
+#if os(FreeBSD)
```
grynspan, line 33: Want to use `.custom()` here @michael-yuji?

PR #323, opened 2025-02-10
Add the necessary swift defines in CMake for Crypto module

cmcgee1024:
### Checklist

- [X] I've run `.script/generate_boilerplate_files_with_gyb` and included updated generated files in a commit of this pull request

### Motivation:

When building SwiftPM with CMake and the swift-crypto package there are errors being produced on macOS relating to missing imports provided by the CryptoKit package.

### Modifications:

Modify the CMakeLists.txt file for the Crypto target so that the defines are set like they are in the Package.swift.

### Result:

After the change, building SwiftPM with CMake on macOS should work.

### DIFF
--- a/Sources/Crypto/CMakeLists.txt
+++ b/Sources/Crypto/CMakeLists.txt
@@ -96,9 +96,13 @@ add_library(Crypto
 target_compile_definitions(Crypto PRIVATE
   "$<$<COMPILE_LANGUAGE:Swift>:CRYPTO_IN_SWIFTPM>")
 
+target_compile_options(Crypto PRIVATE -DCRYPTO_IN_SWIFTPM="true")
+
 if(CMAKE_SYSTEM_NAME STREQUAL "Linux" OR CMAKE_SYSTEM_NAME STREQUAL "Windows" OR CMAKE_SYSTEM_NAME STREQUAL "Android" OR CMAKE_SYSTEM_NAME STREQUAL "WASI" OR CMAKE_SYSTEM_NAME STREQUAL FreeBSD)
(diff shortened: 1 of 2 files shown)


Sources/Crypto/CMakeLists.txt
```diff
@@ -96,9 +96,13 @@ add_library(Crypto
 target_compile_definitions(Crypto PRIVATE
   "$<$<COMPILE_LANGUAGE:Swift>:CRYPTO_IN_SWIFTPM>")
 
+target_compile_options(Crypto PRIVATE -DCRYPTO_IN_SWIFTPM="true")
+
 if(CMAKE_SYSTEM_NAME STREQUAL "Linux" OR CMAKE_SYSTEM_NAME STREQUAL "Windows" OR CMAKE_SYSTEM_NAME STREQUAL "Android" OR CMAKE_SYSTEM_NAME STREQUAL "WASI" OR CMAKE_SYSTEM_NAME STREQUAL FreeBSD)
   target_compile_definitions(Crypto PRIVATE
     "$<$<COMPILE_LANGUAGE:Swift>:CRYPTO_IN_SWIFTPM_FORCE_BUILD_API>")
+
+  target_compile_options(Crypto PRIVATE -DCRYPTO_IN_SWIFTPM_FORCE_BUILD_API="true")
```
Lukasa, line 105: I think these are in the wrong place, they're being set above. The issue is you're missing them in _CryptoExtras.

cmcgee1024: Confirmed. This doesn't seem to be fixing the problem.

Lukasa (approved):

PR #331, opened 2025-02-28
Replace platforms from Package manifest with availability guards

gjcairo:
Minimum deployment versions in package manifests can be problematic for libraries, because changing them or adding them to an existing package that doesn't already use them forces SemVer breaking majors. 

Furthermore, depending on a new library that specifies minimum deployment targets would force the adopting library to also have to add these platforms to its manifest (which would in turn force a major).

This PR removes the platforms stanza from the package manifest and replaces them with availability guards.
This will allow more packages to use swift-crypto, without forcing them to mint new majors to support the appropriate platforms.


Lukasa (commented):
Let's undo the change in crypto-shasum and instead have that done manually. The tool has gone a little haywire there.

Lukasa (approved):
Nice one, thanks @gjcairo!

Issue #338, opened 2025-03-12
'ARC' is only available in iOS 13.2 or newer


bddq:
### New Issue Checklist

- [x] I'm using the latest version of Swift Crypto (`main` branch)
- [x] I read the [Contribution Guidelines](https://github.com/apple/swift-crypto/blob/main/CONTRIBUTING.md)
- [x] I searched for [existing GitHub issues](https://github.com/apple/swift-crypto/issues)

### Expected behavior
Cannot build CryptoExtras because extension is available from iOS 13.0 but enum is available from iOS 13.2.

### Actual behavior
I got an error while building:
`'ARC' is only available in iOS 13.2 or newer`
on ARCEncoding.swift
```
@available(macOS 10.15, iOS 13, watchOS 6, tvOS 13, *)
extension ARC.PresentationState {
```

### Steps to reproduce

Try to build a project that embed `_CryptoExtras`target.

### Swift Crypto version/commit hash

3.12.1

### Environment

<details>
  <pre>Swift version: Apple Swift version 6.0.3 (swiftlang-6.0.3.1.10 clang-1600.0.30.1)
Target: arm64-apple-macosx15.0
Unix version: Darwin donald.ville 24.3.0 Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:24 PST 2025
 root:xnu-11215.81.4~3/RELEASE_ARM64_T6030 arm64

macOS version: 15.3.1
Xcode-select path: '/Applications/Xcode-16.2.0.app/Contents/Developer
Xcode: Xcode 16.2 (Build version 16C5032a)
</pre>
</details>

bddq:
It seems to be related to #334 as the issue is not reproductible in version 3.11.2.

simonjbeaumont:
Thanks @bddq. Looks like we got these annotations slightly off and didn't catch in CI because we don't CI for iOS. We'll try and get a fix out quickly. Sorry for the inconvenience and, in the meantime, please pin back if this is blocking you.

gjcairo:
@bddq thanks for opening this. I've opened https://github.com/apple/swift-crypto/pull/339 with a fix. I still want to check all other platforms build properly before merging it but I need some additional time for that. 

In the meantime, would you mind checking your project builds fine with this patch?

simonjbeaumont:
This should be resolved in https://github.com/apple/swift-crypto/releases/tag/3.12.2. Please do confirm. And a special thank you to @gjcairo for a quick fix here ❤

[...]

PR #339, opened 2025-03-12
Fix availability mismatches

gjcairo:
https://github.com/apple/swift-crypto/issues/338 pointed to some availability guard platform mismatches on `ARC` and its extensions. This PR aims to fix these and make sure all platforms build correctly.


simonjbeaumont:
Thanks for the quick patch here @gjcairo. LGTM assuming it builds on the other platforms too.

gjcairo:
Just validated it builds fine with this patch on watchOS, visionOS, macOS, iOS and tvOS - I'll mark it as ready for review.

simonjbeaumont (approved):
Thank you @gjcairo!

PR #342, opened 2025-03-14
Use centralized CMake checks

rnro:
Use centralized CMake checks

### Motivation:

We now have centrally-hosted CMake checks in the NIO repository which we should use to reduce maintenance. The new checks also increase coverage by performing a build with CMake as part of the checks.

### Modifications:

* Use the new workflow
* Document how to update the CMakeLists files
* Remove the obsolete scripts

### Result:

Better test coverage.


Lukasa (approved):
Nice, LGTM.

CONTRIBUTING.md
```diff
@@ -58,6 +58,20 @@ A good swift-crypto patch is:
 
 Please open a pull request at https://github.com/apple/swift-crypto. Make sure the CI passes, and then wait for code review.
 
+### How to update CMakeLists files
+
+If you wish to update the CMakeLists files you can either do this by using [act](https://github.com/nektos/act) or calling the remotely hosted CI script.
+
+Issuing a command such as the following to `act` will cause it to run the `cmake-lists` job locally which updates the CMakeLists files and if it finds no changes will perform a build. The `--bind` flag means that the VM will be acting on your local filesystem so be aware that this could delete/change files (including your git setup/branches).
```
simonjbeaumont, line 65: > (including your git setup/branches).

OOI, why is this the case (again)? I thought this was mitigated in the github-workflows repo these days?

rnro: Maybe I'm being overly-cautious and I'm wrong. My understanding was that `--bind` just mounts the local directory so if someone writes a workflow with `rm -rf .git` then you're toast.

PR #349, opened 2025-04-05
OpenBSD support.

3405691582:
OpenBSD support.

### Checklist
- [X] I've updated the documentation if necessary (_not necessary_)

- [X] I've run `.script/generate_boilerplate_files_with_gyb` and included updated generated files in a commit of this pull request (_not applicable_)

### Motivation:

swift-crypto is a required dependency for swift-package-manager, and must build and run on OpenBSD.

### Modifications:

    OpenBSD support for cmake/swiftpm.
    
    * Add to patterns for CMakeLists. Also provide a helpful error so future
      porters have a slightly better idea of what might need to be updated.
    
    * Proactively reverse sense in some system name checks so they read as
      "if not Darwin". This is likely safer to do here than elsewhere since
      there are likely not going to be other platforms for which swift-crypto
      already exists.
    
    * Add to Package.swift

    ManagedBuffer.capacity is unavailable on OpenBSD.
    
    ManagedBuffer.capacity depends on malloc introspection (e.g.,
    `malloc_size`), which is not available on this platform -- and is thus
    marked as such.
    
    Here, SecureBytes uses a ManagedBuffer in Backing. We have capacity
    tracked in the BackingHeader, and most of the references to the
    ManagedBuffer.capacity property can be replaced with references to the
    capacity in the header. However, since capacity in the header is marked
    as internal which conflicts with some of the @inline functions, so we
    must change those to @usableFromInline for the platform as well.

    Add conditional for RandomBytes.

    Required for tests to build on OpenBSD.

### Result:

swift-crypto builds and runs on OpenBSD.


Sources/CCryptoBoringSSL/CMakeLists.txt
```diff
@@ -334,6 +334,8 @@ elseif(CMAKE_SYSTEM_NAME MATCHES "Linux|Android|FreeBSD" AND CMAKE_SYSTEM_PROCES
     gen/bcm/vpaes-armv8-linux.S
     gen/crypto/chacha-armv8-linux.S
     gen/crypto/chacha20_poly1305_armv8-linux.S)
+else()
```
Lukasa, line 337: These cmakelists.txt files are autogenerated, so you'll need to modify the generation script and then re-run it to produce this output.

Sources/Crypto/Util/SecureBytes.swift
```diff
@@ -62,7 +62,12 @@ extension SecureBytes {
     @inlinable
     mutating func append<C: Collection>(_ data: C) where C.Element == UInt8 {
         let requiredCapacity = self.count + data.count
-        if !isKnownUniquelyReferenced(&self.backing) || requiredCapacity > self.backing.capacity {
+#if os(OpenBSD)
+        let backingCapacity = self.backing.header.capacity
+#else
+        let backingCapacity = self.backing.capacity
```
Lukasa, line 68: Is there any reason we can't offer a computed property in the OpenBSD case that can offer the `capacity`? That would minimise the scope of this diff, which would help.

Lukasa (changes requested):

Sources/CCryptoBoringSSL/CMakeLists.txt
3405691582: I don't understand. The scripts don't seem to modify the CMakeLists, only the underlying source, which I'm not touching. If it helps, I can leave off the extra helpful error here. Could you elaborate a bit more?

Sources/Crypto/Util/SecureBytes.swift
3405691582: I originally did this because I was slightly concerned about the slight performance impact, but I've added a new commit that adds a computed property (so in the worst case, that commit can be reverted on its own without breaking swift-crypto for the platform). 

This necessitates redefining `_copyBytes` and `_withVeryUnsafeMutableBytes` as `@usableFromInline` instead of `@inlinable` on all platforms.

Sources/Crypto/Util/SecureBytes.swift
```diff
@@ -327,6 +330,15 @@ extension SecureBytes {
             }
         }
 
+        @usableFromInline
+	var allocatedCapacity: Int {
```
Lukasa, line 334:
```suggestion
        var allocatedCapacity: Int {
```

Sources/Crypto/Util/SecureBytes.swift
```diff
@@ -458,9 +470,9 @@ extension SecureBytes.Backing: ContiguousBytes {
     }
 
     /// Very unsafe in the sense that this points to uninitialized memory. Used only for implementations within this file.
-    @inlinable
-    /* private but inlinable */ func _withVeryUnsafeMutableBytes<T>(_ body: (UnsafeMutableRawBufferPointer) throws -> T) rethrows -> T {
-        let capacity = self.capacity
+    @usableFromInline
```
Lukasa, line 473: Why were these dropped from `@inlinable`?

Sources/Crypto/Util/SecureBytes.swift
3405691582: I think I missed a push here which fixed this formatting error. This should be fixed now.

Sources/Crypto/Util/SecureBytes.swift
3405691582: I had to change these from `@inlinable` to `@usableFromInline` because the new `allocatedCapacity` computed property refers to `BackingHeader` which is internal; otherwise I would get a compiler error. If you mean the comments, then I missed there was a similar commenting convention for "private but usableFromInline" and added that back in now.

Lukasa: Ah, so let's push the computed property one layer out. We'd like the body of this method to be inlinable, so we can add another computed property in the outer type that is `@usableFromInline`, which can keep this method accessible.

3405691582: I'm not sure I follow, sorry. `_withVeryUnsafeMutableBytes` and `_copyBytes` are extensions on `SecureBytes.Backing` and right now `allocatedCapacity` is defined on `SecureBytes.Backing`.  If I set `_copyBytes` and `_withVeryUnsafeMutableBytes` back to `@inlinable` and add an extra computed property on `SecureBacking`, I still get errors, so I think I'm misunderstanding something.

Lukasa: Hrm, if the two methods are on `SecureBytes.Backing` and `allocatedCapacity` is also on `SecureBytes.Backing`, merely making `allocatedCapacity` `@usableFromInline` should be enough to flip these back to `@inlinable`.

3405691582: The error specifically reads something like this (same for `_withVeryUnsafeMutableBytes`):

```
/tmp/swift-crypto/Sources/Crypto/Util/SecureBytes.swift:443:51: error: property 'allocatedCapacity' is internal and cannot be referenced from an '@inlinable' function
332 | 
333 |         @usableFromInline
334 |         var allocatedCapacity: Int {
    |             `- note: property 'allocatedCapacity' is not '@usableFromInline' or public
335 | #if os(OpenBSD)
336 |                 return self.header.capacity
    :
441 |     /* private but inlinable */ func _copyBytes<C: Collection>(_ bytes: C, at offset: Int) where C.Element == UInt8 {
442 |         precondition(offset >= 0)
443 |         precondition(offset + bytes.count <= self.allocatedCapacity)
    |                                                   `- error: property 'allocatedCapacity' is internal and cannot be referenced from an '@inlinable' function
444 | 
445 |         let byteRange = offset..<(offset + bytes.count)
```

Lukasa: That error seems...entirely wrong. It's `@usableFromInline`, the annotation is right here.

3405691582: I found an apparent workaround: adding the property to the same section that `_copyBytes` is in makes this work somehow.

Lukasa (approved):
Nice one, this is looking good. Thanks!

3405691582:
As a heads-up, I think this works fine with swiftpm/`swift build` but not with CMake (e.g. during swiftpm bootstrap). I'll send a followup pr if I can untangle the knot.

PR #352, opened 2025-04-18
Fix cmake build.

3405691582:
Fix cmake build.

### Checklist
- [X] I've updated the documentation if necessary **(not necessary)

- [X] I've run `.script/generate_boilerplate_files_with_gyb` and included updated generated files in a commit of this pull request **(not necessary)**

### Motivation:

swift-crypto, a swiftpm dependency, is not compiling when bootstrapping swiftpm with cmake. It is however building fine when built with an existing swiftpm.

In #349 we added a new computed property to handle the fact that ManagedBuffer.capacity is unavailable on OpenBSD. In doing this, however we encountered some strange errors about this property being `internal and cannot be referenced from an '@inlinable' function` despite the fact we marked it `@usableFromInline`. Moving the computed property to the nested class extension seemed to help, but in recent versions of swift this caused a different strange compiler error about the computed property having the `wrong linkage`.

### Modifications:

While I don't really understand what is happening here enough to say whether this is a compiler bug or not, this only seems to be triggering from `_copyBytes` and `_withVeryUnsafeMutableBytes` still. We want these to be inlinable, so the next natural workaround seems to be to plumb the capacity references in by argument instead of referencing `self`. This makes the callsites a bit wordy, but it seems to work.

### Result:

swiftpm successfully bootstraps with cmake.

### DIFF
--- a/Sources/Crypto/Util/SecureBytes.swift
+++ b/Sources/Crypto/Util/SecureBytes.swift
@@ -46,7 +46,7 @@ struct SecureBytes {
     @usableFromInline
     init(unsafeUninitializedCapacity: Int, initializingWith callback: (inout UnsafeMutableRawBufferPointer, inout Int) throws -> Void) rethrows {
         self.backing = Backing.create(capacity: unsafeUninitializedCapacity)
-        try self.backing._withVeryUnsafeMutableBytes { veryUnsafePointer in
+        try self.backing._withVeryUnsafeMutableBytes(self.backing.allocatedCapacity) { veryUnsafePointer in
             // As Array does, we want to truncate the initializing pointer to only have the requested size.
             var veryUnsafePointer = UnsafeMutableRawBufferPointer(rebasing: veryUnsafePointer.prefix(unsafeUninitializedCapacity))
             var initializedCount = 0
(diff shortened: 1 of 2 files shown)


Lukasa:
Hmm, before we merge this, do you have the build output from the failing build? I'm a touch curious about what's going on here.

3405691582:
Here's two build log outputs using the April 12 Linux nighlies installed at `/tmp/toolchain` and the build invocation of `cmake -GNinja -DCMAKE_Swift_COMPILER=/tmp/toolchain/usr/bin/swiftc /tmp/swift-crypto/`. The [log1.txt](https://github.com/user-attachments/files/19868981/log1.txt) output has the build output before the patch, and [log2.txt](https://github.com/user-attachments/files/19868987/log2.txt) output has the clean 
build output after the patch. Attached also is [log3.txt](https://github.com/user-attachments/files/19869092/log3.txt) which is the verbose output from `swift build` without this patch: oddly enough, this builds fine here.

Also: I think at HEAD this problem doesn't manifest as a compiler crash and instead has a "wrong linkage" error when you attempt to build, but the crash logs seem to be a bit more informative. This also might be fixed post April 12, but I haven't had a chance to update and double-check yet.

3405691582:
This change still seems to be necessary. Any updates?

3405691582:
Actually, this might only be necessary on the release branches. I'll do some more investigating.

3405691582:
FWIW, this also seems to be necessary for the swift release/6.3 swiftpm bootstrap as well.

PR #358, opened 2025-06-04
Add "external mu" variant of ML-DSA (65 and 87) in CryptoExtras

fpseverino:
Add an "external mu" variant of the ML-DSA API (65 and 87 variants) in `CryptoExtras`.

### Checklist

- [X] I've run `.script/generate_boilerplate_files_with_gyb` and included updated generated files in a commit of this pull request

### Motivation:

BoringSSL now has an “external mu” variant of the ML-DSA API, which makes it possible to calculate the hash (called “mu”) of the message, context and public key separately, and then pass this “mu” to a signing function.
Basically, it is the same process as the traditional signature function, but separated into two steps.
This is supposed to be the alternative to the HashML-DSA algorithm, which does pure pre-hashing.

### Modifications:

Add the "external mu" API to `MLDSA65` and `MLDSA87` when importing `CryptoExtras`, add tests.

### Result:

The "external mu" variant of ML-DSA is now available in `CryptoExtras`.


Sources/Crypto/Signatures/BoringSSL/MLDSA_boring.swift.gyb
```diff
@@ -25,13 +25,6 @@ import Foundation
 // MARK: - Generated file, do NOT edit
 // any edits of this file WILL be overwritten and thus discarded
 // see section `gyb` in `README` for details.
-
-@_implementationOnly import CCryptoBoringSSL
-#if canImport(FoundationEssentials)
-import FoundationEssentials
-#else
-import Foundation
-#endif
```
fpseverino, line 34: These imports were duplicated in the same file

Sources/_CryptoExtras/MLDSA/MLDSA+externalMu.swift
```diff
@@ -0,0 +1,98 @@
...
+extension MLDSA65.PrivateKey {
+    /// Generate a signature for the prehashed message representative (a.k.a. "external mu").
+    ///
+    /// > Note: The message representative should be obtained via calls to ``MLDSA65/PublicKey/prehash(for:context:)``.
+    ///
+    /// - Parameter mu: The prehashed message representative (a.k.a. "external mu").
+    ///
+    /// - Returns: The signature of the prehashed message representative.
+    public func signature(forPrehashedMessageRepresentative mu: some DataProtocol) throws -> Data {
+        try self.signature_boring(forPrehashedMessageRepresentative: mu)
```
fpseverino, line 36: The `public` method here and the equivalent `package` one defined in `Crypto/Signatures/MLDSA.swift` must have different names, otherwise the compiler thinks we are calling the same method and throws the warning `Function call causes an infinite recursion`.

Sources/Crypto/Signatures/MLDSA.swift
```diff
@@ -80,6 +83,34 @@ extension MLDSA65 {
...
+        ///   - context: The context of the message.
+        ///
+        /// - Returns: The prehashed message representative (a.k.a. "external mu").
+        package func prehash_boring<D: DataProtocol, C: DataProtocol>(for data: D, context: C) throws -> Data {
+            try self.boringSSLKey.prehash_boring(for: data, context: context)
+        }
+
+        private var boringSSLKey: OpenSSLMLDSAPublicKeyImpl<MLDSA65> {
+            get throws {
+                self.impl is OpenSSLMLDSAPublicKeyImpl<MLDSA65>
```
fpseverino, line 109: This throws the warning `'is' test is always true`, should we use `#if` checks instead?

fpseverino: I've updated it to use `#if` checks

PR #359, opened 2025-06-09
Bring swift-crypto up to date with CryptoKit 2025 Beta 1

Lukasa:
This PR sets swift-crypto up for alignment with the WWDC 2025 CryptoKit APIs. This includes the parity APIs for MLKEM and MLDSA, as well as XWing.

At this time the SHA3 APIs are disabled, as they require a novel implementation strategy. This will come later in the summer. All API features that require SHA3 are therefore also disabled at runtime.


Sources/_CryptoExtras/KEM/MLKEM+Compatibility.swift
fpseverino, line 1: Is there a planned minor release of `swift-crypto` before this major one? Because otherwise these "Compatibility" files would not be needed.

In any case, I could update the current implementation of MLKEM and MLDSA to use the future CryptoKit API and simplify everything a bit.

fpseverino:
> as well as potentially to lower the availability guards to match the original implementation.

I'm sorry, I didn't quite understand what you mean here. Could you break that down for me?

For ML-DSA, the minimum version was macOS 10.15 (because of `CryptoKitError`) and I see that it is still that.
As for ML-KEM, in my implementation the minimum version was macOS 14, because in CryptoKit the `KEMPrivateKey` and `KEMPublicKey` protocols [require macOS 14](https://developer.apple.com/documentation/cryptokit/kemprivatekey) (though their versions in `swift-crypto` [only require 10.15](https://github.com/apple/swift-crypto/blob/3ef65592d6d7c35b6c5f43c5afcf2dbbcd98934c/Sources/Crypto/CryptoKitErrors.swift#L18)); now I see that the minimum version of ML-KEM has been lowered to 10.15.

fpseverino:
I think the answer to both questions will have to be the same.

> As for a minor release, that's a good question. We could choose to do so, or not. Do you have a preference?

If the minimum required version of ML-DSA and ML-KEM will be raised in September, I don't know how much sense there is in releasing them now with a lower minimum required version, especially considering that many packages already depend on the `"3.0.0"..<"5.0.0"` version range. It could be a breaking change.

> One of the questions I have will be, can we offer a graceful way to opt in to lowering those guards. Is that even worth doing?

On the other hand, it would be nice if these algorithms worked with pre v26 versions, but that would only be possible in Linux and platforms using BoringSSL, right? On Apple platforms it would still require v26, and that would create a discrepancy.

To conclude, in my opinion the simplest and clearest thing would be to release the new post-quantum algorithms with the API and availability guards they will have in CryptoKit.
I just hope that if this is the case, a beta of `swift-crypto` v4 will be released soon, so that we can start playing with it.

cc @ptoffy as a potential downstream customer who could use ML-DSA in JWTKit

Lukasa:
That makes sense @fpseverino, thanks for that proposal. Yes, I'll try to take a beta shortly: I'll clean up the failing CI and then start a side branch for managing the beta releases.

rnro (approved):

PR #373, opened 2025-07-01
Remove security backend

josephnoir:
### Motivation:

Depending on the platform Swift Crypto Extras might use an
implementation backed by `Security.framework`, specifically for RSA. Since
BoringSSL is a necessary dependency on all platforms, we can deduplicate
the implementation and keep a single backend.

### Modifications:

Remove the `Security.framework` implementation and tie the RSA backend to
BoringSSL.

### Result:

Swift Crypto Extras always uses the BoringSSL backend for RSA.


Lukasa (approved):
Love to see a negative diff, great stuff!

PR #376, opened 2025-07-03
CMake: Declare ASN1 dependency

etcwilde:
SwiftCrypto depends on SwiftASN1 which is declared in the exports file, but does not try to find the project, resulting in build failures of downstream projects that fail to explicitly call
`find_package(SwiftASN1)` before `find_package(SwiftCrypto)`.

Generally, it is recommended that project config files use `find_dependency` to search for the dependencies of the project they represent (https://cmake.org/cmake/help/v3.26/module/CMakeFindDependencyMacro.html). If the package is already found, either through `find_package` in the parent project, or through another means, the `find_dependency` call is a no-op.


etcwilde:
CC @compnerd

compnerd (approved):
I think that we should start following this more universally. We haven't done that in the past.

etcwilde:
> I think that we should start following this more universally. We haven't done that in the past.

I agree with that. It would help clean up some of the fun with dispatch and foundation too.

Lukasa (approved):
Thanks for this @etcwilde!

PR #381, opened 2025-07-21
build: speculative fix for Windows ARM64 builds

compnerd:
Add the Windows ARM64 assembly files.


etcwilde:
It looks like the work to update the Swift-NIO script has already been done https://github.com/apple/swift-nio/pull/3275

PR #394, opened 2025-08-04
Add AES CMAC support

Lukasa:
Motivation

AES CMAC is a standardised albeit somewhat uncommon MAC built on top of the AES block cipher. CMAC pops up in a few places, mostly where it is convenient to assume that you have access to AES, but not to a hash function. For example, AES-SIV is built on CMAC.

Modifications

Add CMAC support
Add tests and test vectors

Result

CMAC is supported

### DIFF
--- a/Sources/_CryptoExtras/AES/CMAC.swift
+++ b/Sources/_CryptoExtras/AES/CMAC.swift
@@ -22,7 +22,12 @@ import Foundation
 
 @available(iOS 13.0, macOS 10.15, watchOS 6.0, tvOS 13.0, *)
 extension AES {
+    /// A cipher-based message authentication code.
+    /// 
+    /// CMAC uses AES to implement a MAC.  CMAC is useful in contexts where access to
+    /// a hash function is not guaranteed, but a block cipher will be available.
     public struct CMAC: @unchecked Sendable {
+        // Unchecked sendable because this is CoW.
         fileprivate var backing: Backing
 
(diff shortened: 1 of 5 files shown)


Sources/_CryptoExtras/AES/CMAC.swift
```diff
@@ -0,0 +1,195 @@
...
+
+#if canImport(FoundationEssentials)
+import FoundationEssentials
+#else
+import Foundation
+#endif
+
+@available(iOS 13.0, macOS 10.15, watchOS 6.0, tvOS 13.0, *)
+extension AES {
+    public struct CMAC: @unchecked Sendable {
```
glbrntt, line 25: Can you add a "why `@unchecked` is okay here" comment?

Sources/_CryptoExtras/AES/CMAC.swift
```diff
@@ -0,0 +1,195 @@
...
+            }
+
+            self.backing = Backing(key: key, outputSize: outputSize)
+        }
+
+        /// Adds data to be authenticated by MAC function. This can be called one or more times to append additional data.
+        ///
+        /// - Parameters:
+        ///   - data: The data to be authenticated.
+        public mutating func update(bufferPointer: UnsafeRawBufferPointer) {
```
glbrntt, line 58: The labels don't line up here

glbrntt (approved):
Two tiny things but LGTM otherwise

Issue #223, opened 2024-03-15
Linux armv7 support


orobio:
On linux/armv7 some symbols are not properly prefixed. Building a project that includes both swift-nio-ssl and swift-crypto, results in the following conflicts:
```text
ld.lld: error: duplicate symbol: CRYPTO_poly1305_init_neon
ld.lld: error: duplicate symbol: CRYPTO_poly1305_update_neon
ld.lld: error: duplicate symbol: CRYPTO_poly1305_finish_neon
ld.lld: error: duplicate symbol: openssl_poly1305_neon2_blocks
ld.lld: error: duplicate symbol: openssl_poly1305_neon2_addmulmod
ld.lld: error: duplicate symbol: vpaes_encrypt_key_to_bsaes
ld.lld: error: duplicate symbol: vpaes_decrypt_key_to_bsaes
ld.lld: error: duplicate symbol: sha512_block_data_order_neon
ld.lld: error: duplicate symbol: sha256_block_data_order_neon
ld.lld: error: duplicate symbol: bsaes_cbc_encrypt
ld.lld: error: duplicate symbol: bsaes_ctr32_encrypt_blocks
ld.lld: error: duplicate symbol: x25519_NEON
ld.lld: error: duplicate symbol: CRYPTO_has_broken_NEON
ld.lld: error: duplicate symbol: CRYPTO_needs_hwcap2_workaround
```

It looks like [CCryptoBoringSSL_boringssl_prefix_symbols.h](https://github.com/apple/swift-crypto/blob/main/Sources/CCryptoBoringSSL/include/CCryptoBoringSSL_boringssl_prefix_symbols.h) is generated by [vendor-boringssl.sh](https://github.com/apple/swift-crypto/blob/main/scripts/vendor-boringssl.sh) by building BoringSSL for various architectures and gathering all symbols. There is no official linux/armv7 docker image yet, so at the moment we can't use the same method for adding these symbols, but would it be possible to create a temporary solution, for example by hardcoding these symbols? Or is there any proper solution that could work without requiring an official linux/armv7 docker?

Lukasa:
I'd be open to an _unofficial_ docker image, but I don't think hardcoding the symbols themselves is going to work. Another option might be to produce [a cross-compilation SDK](https://github.com/apple/swift-evolution/blob/main/proposals/0387-cross-compilation-destinations.md) that supports an armv7 toolchain, which we could then integrate. @MaxDesiatov might be able to help with that.

[...]

PR #396, opened 2025-08-10
Improve vendor-boringssl.sh script to make it work better

xtremekforever:
Updates for the `scripts/vendor-boringssl.h` script that make it a bit easier to use and ensure you don't accidentally get an update of BoringSSL you don't want.

### Motivation:

I was working on investigating building a Swift SDK for #223 and ran into a couple of things that I saw could be improved. Not all need to be done, but at the very least I feel the boringssl revision lock should be considered. @Lukasa 

### Modifications:

- Add missing invocation of `$sed` in the script- `gsed` was being invoked instead, although everywhere else `$sed` was used (typo).
- Support locking `BORINGSSL_REVISION` to a specific revision number through an argument to the script. This way the script can be re-run without unintentionally upgrading boringssl.
- Remove the non-existent `boringssl_prefix_symbols_nasm.inc` exclude from the `CCryptoBoringSSL` target. I'm guessing it was there for an older version of boringssl, but no longer there with the current revision?

### Result:

The `scripts/vendor_boringssl.sh` script behaves a bit better and can be run without upgrading the boringssl version unless explicitly desired.

### DIFF
--- a/scripts/vendor-boringssl.sh
+++ b/scripts/vendor-boringssl.sh
@@ -101,9 +101,9 @@ function mangle_symbols {
         )
 
         # Now cross compile for our targets.
-        docker run -t -i --rm --privileged -v"$(pwd)":/src -w/src --platform linux/arm64 swift:5.10-jammy \
+        docker run -t -i --rm --privileged -v"$(pwd)":/src -w/src --platform linux/arm64 swift:6.0-jammy \
             swift build --product CCryptoBoringSSL
-        docker run -t -i --rm --privileged -v"$(pwd)":/src -w/src --platform linux/amd64 swift:5.10-jammy \
+        docker run -t -i --rm --privileged -v"$(pwd)":/src -w/src --platform linux/amd64 swift:6.0-jammy \
             swift build --product CCryptoBoringSSL
 
         # Now we need to generate symbol mangles for Linux. We can do this in
--- a/scripts/vendor-boringssl.sh
+++ b/scripts/vendor-boringssl.sh
@@ -248,7 +248,7 @@ echo "DISABLING assembly on x86 Windows"
     # x86 Windows builds require nasm for acceleration. SwiftPM can't do that right now,
     # so we disable the assembly.
     cd "$DSTROOT"
-    gsed -i "/#define OPENSSL_HEADER_BASE_H/a#if defined(_WIN32) && (defined(__x86_64) || defined(_M_AMD64) || defined(_M_X64) || defined(__x86) || defined(__i386) || defined(__i386__) || defined(_M_IX86))\n#define OPENSSL_NO_ASM\n#endif" "include/openssl/base.h"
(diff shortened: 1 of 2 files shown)


scripts/vendor-boringssl.sh
```diff
@@ -101,9 +101,9 @@ function mangle_symbols {
         )
 
         # Now cross compile for our targets.
-        docker run -t -i --rm --privileged -v"$(pwd)":/src -w/src --platform linux/arm64 swift:5.10-jammy \
+        docker run -t -i --rm --privileged -v"$(pwd)":/src -w/src --platform linux/arm64 swift:6.0-jammy \
             swift build --product CCryptoBoringSSL
-        docker run -t -i --rm --privileged -v"$(pwd)":/src -w/src --platform linux/amd64 swift:5.10-jammy \
+        docker run -t -i --rm --privileged -v"$(pwd)":/src -w/src --platform linux/amd64 swift:6.0-jammy \
```
Lukasa, line 106: We should usually keep this on the oldest supported Swift compiler, which for now remains 5.10. It will become 6.0 when 6.2 goes GA.

scripts/vendor-boringssl.sh
```diff
@@ -170,7 +170,8 @@ echo "CLONING boringssl"
 mkdir -p "$SRCROOT"
 git clone https://boringssl.googlesource.com/boringssl "$SRCROOT"
 cd "$SRCROOT"
-BORINGSSL_REVISION=$(git rev-parse HEAD)
+BORINGSSL_REVISION=035e720641f385e82c72b7b0a9e1d89e58cb5ed5
+git checkout $BORINGSSL_REVISION
```
Lukasa, line 174: I am inclined to flip this behaviour and say that we should enable the passing of an optional flag to the script that is a revision and, if passed, we should check that out. That way we don't end up encoding the hash in the script itself, but enable optional repeatable runs.

xtremekforever: I can do that!! Anything that allows us to lock the version when just maintaining and updating other pieces of the script.

xtremekforever: So now the revision can be passed to the script like this:

```console
$ ./scripts/vendor-boringssl.sh 035e720641f385e82c72b7b0a9e1d89e58cb5ed5
REMOVING any previously-vendored BoringSSL code
CLONING boringssl
Cloning into '/tmp/.workingN8YfSl/src/boringssl.googlesource.com/boringssl'...
remote: Finding sources: 100% (25/25)
remote: Total 124009 (delta 74986), reused 124005 (delta 74986)
Receiving objects: 100% (124009/124009), 289.25 MiB | 77.32 MiB/s, done.
Resolving deltas: 100% (74986/74986), done.
Updating files: 100% (7939/7939), done.
CHECKING OUT boringssl@035e720641f385e82c72b7b0a9e1d89e58cb5ed5
Note: switching to '035e720641f385e82c72b7b0a9e1d89e58cb5ed5'.
```

As you can see, it will give a message about using that revision before running the `git checkout` command.

xtremekforever:
> Looks like shellcheck has a few concerns.

I've updated the script with the quotes as requested. Also noticed that if an argument was not passed the script would give this error:

```
./scripts/vendor-boringssl.sh: line 49: $1: unbound variable
```

And fixed this one as well!

Lukasa:
Thanks!

PR #412, opened 2025-09-17
Remove useless `try` to fix warnings

ptoffy:
`getKeyPrimitives` is marked as throws but is not actually throwing errors and this is yielding warnings. This fixes the warnings.

### Checklist

- [N/A] I've run `./scripts/generate_boilerplate_files_with_gyb.sh` and included updated generated files in a commit of this pull request

### Motivation:

Currently there's a warning
```
/swift-crypto/Sources/_CryptoExtras/RSA/RSA_boring.swift:67:9: warning: no calls to throwing functions occur within 'try' expression
```
because we're using `try` on a non throwing function. This escalates to all of the functions that use it. 
I'm not sure if removing `throws` from a non-actually-throwing public function is breaking, as using `try` on a non throwing function only yields a warning (if not in warnings-as-errors mode) but it actively is a signature change so it might still be breaking. In case we don't want to merge this as is, the warning can be removed by simply removing useless `try`s and we can wait for the next major version to remove the false-throwers.

### Modifications:

Remove the useless `try` and `throws` where they're not needed.

### Result:

The warning is gone.

### DIFF
--- a/Sources/_CryptoExtras/RSA/RSA.swift
+++ b/Sources/_CryptoExtras/RSA/RSA.swift
@@ -141,7 +141,7 @@ extension _RSA.Signing {
             self.backing = backing
         }
 
-        public func getKeyPrimitives() -> Primitives {
+        public func getKeyPrimitives() throws -> Primitives {
             let (n, e) = self.backing.getKeyPrimitives()
             return Primitives(modulus: n, publicExponent: e)
         }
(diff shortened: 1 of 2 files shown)


Sources/_CryptoExtras/RSA/RSA.swift
```diff
@@ -538,8 +538,8 @@ extension _RSA.Encryption {
         public var keySizeInBits: Int { self.backing.keySizeInBits }
         fileprivate init(_ backing: BackingPublicKey) { self.backing = backing }
 
-        public func getKeyPrimitives() throws -> Primitives {
-            let (n, e) = try self.backing.getKeyPrimitives()
+        public func getKeyPrimitives() -> Primitives {
```
Lukasa, line 541: Let's leave the `throws` on the public functions and just remove the incorrect internal `try`s.

Lukasa:
Thanks @ptoffy!

PR #423, opened 2025-10-28
Incorrect maximum encryption size for RSA OAEP

mschrag:
Incorrect maximum encryption size for RSA OAEP

### Checklist

- [n/a] I've run `./scripts/generate_boilerplate_files_with_gyb.sh` and included updated generated files in a commit of this pull request

### Motivation:
The `maximumEncryptSize` function is hardcoded to use 42 as the hash offset, but the RFC actually says it's "2*hLen-2" so 42 is only valid for SHA1. SHA256 should be 62 (2*32-2). This adds a hash length onto the Digest enum, which can then be used in the length calculation.

In writing the tests for this, I also stumbled on the unsafe PEM representation for RSA Public Keys don't allow 1024-bit keys as documented, so this also fixes that.

### Modifications:
* Added a `hashBitLength` to the RSA Digest which is then used in the `maximumEncryptSize` to properly compute the maximum length
* Corrected the minimum key size for RSA unsafe PEM public keys to 1024 from 2048

### Result:
* The `maximumEncryptSize` function will return the expected value for RSA OAEP SHA256 keys. I don't believe anything calls this internally, so this would be for external consumers.
* Unsafe construction of RSA public keys now allow 1024 bit keys. It was documented as supporting them, but the check was still 2048 (probably copy-paste error from the safe variant). This only expands the potential uses, so it shouldn't introduce any new failures.

### DIFF
--- a/Sources/CryptoExtras/RSA/RSA.swift
+++ b/Sources/CryptoExtras/RSA/RSA.swift
@@ -508,7 +508,7 @@ extension _RSA.Encryption {
         /// - Warning: Key sizes less than 2048 are not recommended and should only be used for compatibility reasons.
         public init(unsafePEMRepresentation pemRepresentation: String) throws {
             self.backing = try BackingPublicKey(pemRepresentation: pemRepresentation)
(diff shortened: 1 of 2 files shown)


Lukasa (approved):
Nice one, thanks!

PR #436, opened 2026-04-14
Add module anchor to fix empty dynamic framework on Xcode 26

### LINKED ISSUE #435: Crypto target fails to build as dynamic framework on Xcode 26 (empty re-export module produces no Mach-O binary)
## Description

When the `Crypto` product is built as a **dynamic framework** by Xcode 26 (tested through 26.4), the framework directory is created but **no Mach-O binary is produced**, causing a linker error:

```
clang: error: no such file or directory:
  '.../PackageFrameworks/Crypto_17A3B1FFC41E47_PackageProduct.framework/Crypto_17A3B1FFC41E47_PackageProduct'
```

## Root cause

On Apple platforms, all 93 Swift source files in the `Crypto` target compile down to just `@_exported import CryptoKit` (because `CRYPTO_IN_SWIFTPM` is defined but `CRYPTO_IN_SWIFTPM_FORCE_BUILD_API` is not). The resulting module has **zero exported symbols of its own**.

Xcode 26 changed how SPM packages are built for test targets that declare their own `packageProductDependency`: each package product in the dependency graph is built as a separate dynamic framework. A module with no symbols produces a framework directory but no binary — and the linker fails trying to link it.

## How to reproduce

1. Create an Xcode project with a test target
2. Add a SPM package that transitively depends on `swift-crypto` (e.g., `swift-certificates` → `swift-crypto`)
3. Add that package as a `packageProductDependency` on the **test target** (not the app target)
4. Build for testing on Xcode 26 targeting iOS Simulator

The build fails with the error above.

## Why it works on Xcode 16

On Xcode 16 and earlier, SPM packages for test targets were either statically linked or shared with the host app's already-built packages. The empty `Crypto` module was never built as a standalone dynamic framework, so the absence of a binary wasn't a problem.

## Why it works with local packages

When the consuming package is added as a **local package** in an Xcode workspace, Xcode uses static linking and merges everything into the final binary. No separate framework is needed, so the empty module works fine.

## Environment

- Xcode 26.0 through 26.4
- swift-crypto 3.x and 4.x (tested 3.15.1 and 4.3.1)
- iOS Simulator destination
- macOS 26 (Darwin 25.2.0)

## Possible fix

The `Crypto` target on Apple platforms could include at least one non-empty symbol (e.g., a version constant) to ensure the framework binary is always produced, regardless of how Xcode chooses to link it.

mrcararia:
## Summary

Fixes #435.

On Apple platforms every source file in the `Crypto` target compiles to just `@_exported import CryptoKit` (because `CRYPTO_IN_SWIFTPM` is defined but `CRYPTO_IN_SWIFTPM_FORCE_BUILD_API` is not), producing **zero exported symbols**.

Xcode 26 changed how SPM packages are built for test-target dependencies: each package product is built as a separate dynamic framework. A module with no symbols generates a `.framework` directory but no Mach-O binary, causing a linker error:

```
clang: error: no such file or directory:
  '.../PackageFrameworks/Crypto_17A3B1FFC41E47_PackageProduct.framework/Crypto_17A3B1FFC41E47_PackageProduct'
```

## Fix

Added a minimal `_CryptoModuleAnchor` public enum (outside the conditional compilation guard) so the module always produces at least one exported symbol, regardless of how it is linked. The type is prefixed with `_` to signal it is not intended for public use.

## Test plan

- [ ] Build a project with a test target that has `swift-crypto` (or a transitive dependency like `swift-certificates`) as a `packageProductDependency` on Xcode 26 targeting iOS Simulator — should now succeed
- [ ] Verify existing tests still pass (`swift test`)


Lukasa (changes requested):
I think we can achieve this by providing a single func as a `@usableFromInline` or `_@spi` symbol which better achieves the outcome. Please also put this file into a `BoringSSL` directory, to ensure it is preserved across updates.

mrcararia:
Thanks for the review @Lukasa! Addressed both points in ee838cf:

- Replaced the `public enum _CryptoModuleAnchor` with a single `@usableFromInline internal func _cryptoModuleAnchor()` so the anchor no longer shows up in the public API surface.
- Moved the file to `Sources/Crypto/Util/BoringSSL/_CryptoModuleAnchor.swift` so it is preserved across upstream syncs. Happy to relocate it elsewhere if you prefer a different `BoringSSL/` directory.

Lukasa (approved):

PR #450, opened 2026-07-28
Support building on platforms without vendored assembly (OPENSSL_NO_ASM fallback, powerpc64le)

colemancda:
Allow swift-crypto to build on Linux architectures that have no vendored BoringSSL assembly, and restore target detection for powerpc64le.

### Checklist

### Motivation:

The CMake build only defines platform sources for x86_64 and arm64 (plus their Windows variants); every other architecture — 32-bit ARM, i686, powerpc64le, RISC-V, MIPS — hits a hard `message(FATAL_ERROR "platform sources are not defined here ...")` with no way to opt into a portable build, even though BoringSSL fully supports a pure-C configuration via `OPENSSL_NO_ASM` (the assembly is only an optimization guarded by `!defined(OPENSSL_NO_ASM)`).

Additionally, upstream BoringSSL removed powerpc64le from its target detection, so on ppc64le neither `OPENSSL_32_BIT` nor `OPENSSL_64_BIT` gets defined and every header fails with:

```
CCryptoBoringSSL_bn.h:173:2: error: "Must define either OPENSSL_32_BIT or OPENSSL_64_BIT"
```

The comment above that `#error` explicitly says the portable C code works on any standard 32/64-bit two's-complement little-endian architecture and invites carrying a local patch for such targets — powerpc64le qualifies.

We hit both issues cross-compiling swift-crypto with CMake for embedded Linux targets (Buildroot) across armv5/armv6/armv7/i686/aarch64/x86_64/ppc64le; with these two changes the full matrix builds and the `Crypto` module passes a runtime smoke test (`SHA256.hash`) on the non-asm architectures.

### Modifications:

- `Sources/CCryptoBoringSSL/CMakeLists.txt`: replace the `FATAL_ERROR` fallback branch with `target_compile_definitions(CCryptoBoringSSL PRIVATE OPENSSL_NO_ASM)`. Platforms that do have vendored assembly are unaffected.
- `Sources/CCryptoBoringSSL/include/CCryptoBoringSSL_target.h`: add a `__powerpc64__` + little-endian branch defining `OPENSSL_64_BIT`. Since this file is vendored, the change is carried as `scripts/patch-4-powerpc64le-target.patch` and applied by `scripts/vendor-boringssl.sh` (alongside the existing patch-1/2) so it survives re-vendoring. `OPENSSL_PPC64LE` is deliberately not defined so no stale arch-specific code paths are enabled; the generic C implementation is used together with `OPENSSL_NO_ASM`.

### Result:

The CMake build succeeds on Linux architectures without vendored assembly (32-bit ARM, i686, powerpc64le, ...) using BoringSSL's supported pure-C configuration, instead of failing at configure time. x86_64 and arm64 continue to build their vendored assembly exactly as before. The powerpc64le detection fix also benefits SwiftPM builds on that architecture.

### DIFF
--- a/Sources/CCryptoBoringSSL/CMakeLists.txt
+++ b/Sources/CCryptoBoringSSL/CMakeLists.txt
@@ -361,7 +361,10 @@ elseif(CMAKE_SYSTEM_NAME MATCHES "Windows" AND CMAKE_SYSTEM_PROCESSOR MATCHES "A
     gen/crypto/chacha-armv8-win.S
     gen/crypto/chacha20_poly1305_armv8-win.S)
 else()
-   message(FATAL_ERROR "platform sources are not defined here for ${CMAKE_SYSTEM_NAME} on ${CMAKE_SYSTEM_PROCESSOR}")
+  # No assembly is vendored for this platform. Fall back to the portable
+  # pure-C implementation; the assembly is only an optimization guarded by
+  # !defined(OPENSSL_NO_ASM).
+  target_compile_definitions(CCryptoBoringSSL PRIVATE OPENSSL_NO_ASM)
 endif()
 
 target_include_directories(CCryptoBoringSSL PUBLIC
--- a/Sources/CCryptoBoringSSL/include/CCryptoBoringSSL_target.h
+++ b/Sources/CCryptoBoringSSL/include/CCryptoBoringSSL_target.h
@@ -54,6 +54,11 @@
 #define OPENSSL_32_BIT
 #elif defined(__myriad2__)
 #define OPENSSL_32_BIT
+#elif defined(__powerpc64__) && defined(__BYTE_ORDER__) && \
+    __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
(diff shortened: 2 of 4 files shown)


simonjbeaumont:
@colemancda thanks for taking the time to open the PR, however we don't typically want to maintain patches of BoringSSL in Swift Crypto.

The header file being patched includes this note:

```c
// However, we know that anything that seems to work will soon be expected
// to work and, quickly, the implicit expectation is that every machine will
// always work. So this list serves to mark the boundary of what we guarantee.
// Of course, you can run the code any many more machines, but then you're
// taking on the burden of fixing it and, if you're doing that, then you must
// be able to carry local patches. In which case patching this list is trivial.
```

By its absence from the list, the platform is not guaranteed to work by the BoringSSL maintainers. Us patching here shifts that burden to Swift Crypto and we have no way to verify it works today, nor that it will continue to work going forward.

I think your two options are (1) maintain a patched version of Swift Crypto, for your own use case; or (2) try to get this patch upstreamed to BoringSSL itself.

If you were successful with (2), Swift Crypto could be updated to a new version of BoringSSL with the platform support you require.

PR #452, opened 2026-08-05
[5.x] Adopt CheckImplementationOnly, cleanup ABI, and be warnings-clean

simonjbeaumont:
## Motivation

We've long since used `@_implementationOnly import` (IOI) for our vendored C modules (BoringSSL and libXKCP) in an attempt to not leak our dependency into downstream builds. This is important because leaking might cause symbol collisions (build failures) or, worse, layout mismatches (latent runtime failures).

IOI was intended to be used with library evolution, which this project does not support, and use of IOI without library evolution has been soft-deprecated. Users are encouraged to move to `internal import` (II), which is usually the right choice for Swift packages. But II does not provide the same guarantees as IOI and, for this project, we really need the protections given by IOI so we will continue to use it, even with library evolution disabled.

One of the reasons for the soft-deprecation was the implementation of IOI used an incomplete checker, that won't catch all interface leaks without library evolution. However, Swift has also introduced a new feature, `CheckImplementationOnly` (CIO), which enables the stricter interface checking that we need.

Enabling CIO for this project today actually exposes a number of places where we are leaking the C types into our interface in ways that could cause build- or runtime problems in downstream projects, and rightly causes the build to fail. These types will need to have their C struct members boxed behind an `OpaquePointer`.

## Modifications

This patch enables CIO, and adds the required boxing to the offending types.

The public API is unchanged.

Additional refactoring is also done to ensure a single designated initializer, responsible for the allocation, forcing all convenience initializers through an `init(takingOwnershipOf:)` spelling, to reduce the chances of introducing future double free bugs in complex initializers.

This patch also adds some benchmarks for some of the paths that would be implicated by these changes. While they show the expected increase in allocations, there's only a modest-to-negligible performance regression, since the majority of the compute is dominated by the time spent doing the crypto routines in the C library themselves.

Finally, as we're approaching a new major, it's a good opportunity to get warnings-clean and lock in warnings-as-errors, so this PR does that, suppressing only `ImplementationOnlyDeprecated`, which we cannot eliminate. So we now build in CI using:

    -Xswiftc -warnings-as-errors -Xswiftc -Wwarning -Xswiftc ImplementationOnlyDeprecated

However, Swift 6.4 introduces `SourceWarningControl`, which we can use to actually _silence_ these warnings, using `@diagnose(_:as:)`, which we do conditionally:

    #if hasFeature(SourceWarningControl)
    @diagnose(ImplementationOnlyDeprecated, as: ignored)     @_implementationOnly import ...
    #else
    @_implementationOnly import ...
    #endif

When we raise the minimum supported Swift compiler version to one that supports `SourceWarningControl`, we can remove the `#if hasFeature` guards.

We've had to drop to `--build-system native` (vs swift-build) for Swift 6.4 CI because SwiftPM is not stripping the warning-related compiler flags when building dependencies.

## Result

- ABI no longer leaks C dependencies from BoringSSL and libXKCP and project builds cleanly with `CheckImplementationOnly`.
- All other compiler warnings have been addressed and project builds with `-warnings-as-errors -Wwarning ImplementationOnlyDeprecated` in CI.
- On Swift 6.4+, `ImplementationOnlyDeprecated` warnings are completely silenced and the project is completely warnings-free.


simonjbeaumont:
@Lukasa this is ready for review now, when you have a minute.

Most of the commits here are procedural and uncontroversial to get to a warnings-clean state.

The one of most interest is this one https://github.com/apple/swift-crypto/pull/452/commits/cd73729a558346865cf84c1cd5151a9c5d4c4918, which is the CheckImplementationOnly and boxing work.

simonjbeaumont:
@Lukasa as we discussed offline, it looks like there's some significant difference between `CheckImplementationOnly` between Swift 6.3 and 6.4. It's an experimental feature, so maybe it's expected.

I took the commit that enables it and does the boxing and split it up into all the types we boxed and attempted to revert them one at a time. They're all needed to be CIO-clean on 6.3 and none are needed on 6.4.

To show this in CI I've pushed the branch now with them all reverted and we can see that the build fails on 6.3 but passes on nightly-next (6.4, ([link](https://github.com/apple/swift-crypto/actions/runs/31105222059/job/92628517976?pr=452))). There are many failures on 6.3, but here's an extract:

> ```
> /swift-crypto/Sources/CryptoBoringWrapper/AEAD/BoringSSLAEAD.swift:47:30: error: cannot use type alias 'EVP_AEAD_CTX' in a property declaration marked public or in a '@frozen' or '@usableFromInline' context; 'CCryptoBoringSSL' has been imported as implementation-only
>  45 |     @available(macOS 10.15, iOS 13, watchOS 6, tvOS 13, macCatalyst 13, visionOS 1.0, *)
>  46 |     public class AEADContext {
>  47 |         private var context: EVP_AEAD_CTX
>     |                              `- error: cannot use type alias 'EVP_AEAD_CTX' in a property declaration marked public or in a '@frozen' or '@usableFromInline' context; 'CCryptoBoringSSL' has been imported as implementation-only
>  48 | 
>  49 |         public init<Key: ContiguousBytes>(cipher: BoringSSLAEAD, key: Key) throws {
> ```
-- source: https://github.com/apple/swift-crypto/actions/runs/31105222059/job/92628518043?pr=452#step:6:484

I'll now reapply all the commits and we'll see both 6.3 and 6.4 pass. I'm confirming the state of the feature with the Swift team, but it's possible that we should only be enabling this on 6.4+ and we are already CIO-clean with no changes required.

simonjbeaumont:
Closing this one in favour of a cleaned up one that doesn't do all the unnecessary boxing to satisfy 6.3's false positives. Leaving this one untouched as a potentially useful artifact.

PR #453, opened 2026-08-06
[5.x] Align with CryptoKit APIs from Xcode 27 beta 4

simonjbeaumont:
## Motivation

Xcode 27 beta 4 is now available[^1] with so we can tag a compatible release with the cross-platform API here.

## Modifications

- Update shared API surface to align with CryptoKit

## Result

- Aligned with CryptoKit APIs in latest Xcode beta release.

[^1]: https://developer.apple.com/download/all/?q=Xcode%2027


simonjbeaumont:
```
1 breaking change detected in Crypto:
  💔 API breakage: func ChaChaPoly.open(inplace:using:nonce:tag:authenticating:) has been removed
```

This one is expected. We removed some temporary overloads that were SPI in this sync.

Lukasa (approved):

PR #454, opened 2026-08-07
[5.x] Adopt CheckImplementationOnly and CI with -warnings-as-errors

simonjbeaumont:
## Motivation

We've long since used `@_implementationOnly import` (IOI) for our vendored C modules (BoringSSL and libXKCP) in an attempt to not leak our dependency into downstream builds. This is important because leaking might cause symbol collisions (build failures) or, worse, layout mismatches (latent runtime failures).

IOI was intended to be used with library evolution, which this project does not support, and use of IOI without library evolution has been soft-deprecated. Users are encouraged to move to `internal import` (II), which is usually the right choice for Swift packages. But II does not provide the same guarantees as IOI and, for this project, we really need the protections given by IOI so we will continue to use it, even with library evolution disabled.

One of the reasons for the soft-deprecation was the implementation of IOI used an incomplete checker, that won't catch all interface leaks without library evolution. However, Swift has also introduced a new feature, `CheckImplementationOnly` (CIO), which enables the stricter interface checking that we need, which we should adopt.

Aside: This feature is visible in 6.3 but is incomplete and the Swift team only recommend adopting in 6.4+. Indeed, this PR is a cut-back version of #452, which which included a lengthy patch and discussion, boxing many types that were a result of false positives from the `CheckImplementationOnly` feature on 6.3.

## Modifications

This patch enables `CheckImplementationOnly`, when building with 6.4, or newer.

Additionally, as we're approaching a new major, it's a good opportunity to get warnings-clean and lock in warnings-as-errors, so this PR does that, suppressing only `ImplementationOnlyDeprecated`, which we cannot eliminate. So we now build in CI using:

    -Xswiftc -warnings-as-errors -Xswiftc -Wwarning -Xswiftc ImplementationOnlyDeprecated

However, Swift 6.4 introduces `SourceWarningControl`, which we can use to actually _silence_ these warnings, using `@diagnose(_:as:)`, which we do conditionally:

    #if hasFeature(SourceWarningControl)
    @diagnose(ImplementationOnlyDeprecated, as: ignored)     @_implementationOnly import ...
    #else
    @_implementationOnly import ...
    #endif

When we raise the minimum supported Swift compiler version to one that supports `SourceWarningControl`, we can remove the `#if hasFeature` guards.

We've had to drop to `--build-system native` (vs swift-build) for Swift 6.4 CI because SwiftPM is not stripping the warning-related compiler flags when building dependencies.

## Result

- Stricter checks for not leakingC dependencies from BoringSSL and libXKCP -- project builds cleanly with `CheckImplementationOnly`.
- All other compiler warnings have been addressed and project builds with `-warnings-as-errors -Wwarning ImplementationOnlyDeprecated` in CI.
- On Swift 6.4+, `ImplementationOnlyDeprecated` warnings are completely silenced and the project is completely warnings-free.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   soundness:
     name: Soundness
-    uses: swiftlang/github-workflows/.github/workflows/soundness.yml@0.0.7
+    uses: swiftlang/github-workflows/.github/workflows/soundness.yml@0.0.13
     with:
       license_header_check_project_name: "SwiftCrypto"
       docs_check_enabled: false
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -19,28 +19,23 @@ jobs:
     name: Unit tests
     uses: apple/swift-nio/.github/workflows/unit_tests.yml@main
     with:
-      linux_5_10_arguments_override: "--explicit-target-dependency-import-check error"
-      linux_6_1_arguments_override: "--explicit-target-dependency-import-check error"
-      linux_6_2_arguments_override: "--explicit-target-dependency-import-check error"
-      linux_6_3_arguments_override: "--explicit-target-dependency-import-check error"
-      linux_nightly_next_arguments_override: "--explicit-target-dependency-import-check error"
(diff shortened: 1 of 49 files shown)


Lukasa (approved):
Nice, looking good.

PR #463, opened 2026-08-28
Match the documented parameter names to the signatures

darkdi:
_Match the documented parameter names to the signatures in Crypto and CryptoExtras_

### Checklist

### Motivation:

Both `AES.GCM._SIV.open` overloads document a nonce, with the paragraph about keeping it unique for every use of the key, and neither of them takes one -- it arrives inside the sealed box. That paragraph is copied down from `seal`, where it belongs. In a crypto API this seems worse than a typo, since it sends the reader looking for a parameter that is not there. `Cipher.open` carries the same copy. The other four look like plain name drift.

### Modifications:

Dropped the three nonce entries, renamed four entries to the names the signatures use. Comments only, no code touched.

### Result:

The parameter lists match the signatures.

### DIFF
--- a/Sources/CryptoExtras/AES/AES_GCM_SIV.swift
+++ b/Sources/CryptoExtras/AES/AES_GCM_SIV.swift
@@ -63,7 +63,6 @@ extension AES.GCM {
         /// - Parameters:
         ///   - sealedBox: The sealed box to authenticate and decrypt
         ///   - key: An encryption key of 128 or 256 bits
-        ///   - nonce: An Nonce for AES-GCM-SIV encryption. The nonce must be unique for every use of the key to seal data. It can be safely generated with AES.GCM.Nonce().
         ///   - authenticatedData: Data that was authenticated as part of the seal
         /// - Returns: The ciphertext if opening was successful
         /// - Throws: CipherError errors. If the authentication of the sealed box failed, incorrectTag is thrown.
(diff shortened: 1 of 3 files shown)


simonjbeaumont (changes requested):
@darkdi thanks for taking the time to open this PR 🙏 

In general these housekeeping PRs are usually very welcome -- it's good for us to have the doc comments line up properly, especially for public API -- however, because this project shares code with another changes the tradeoffs.


> Files that do not have the _boring or _xkcp suffix are part of the public API of CryptoKit. Changing these requires passing a higher bar, as any change in these files must be accompanied by a change in CryptoKit itself.

-- source: https://github.com/apple/swift-crypto#code-organisation

Could you revert the changes made to the files in `Crypto/` so that the patch only affects the changes in `CryptoExtras/`?

There's no user benefit to those changes anyway, since those documentation comments are not on public symbols. Conversely, one of the changes you made in `CryptoExtras` is useful since it fixes the broken documentation on public API 👍

darkdi:
Reverted, the patch touches only CryptoExtras now. I missed the code organisation note in the README about the files without the `_boring` suffix, that is on me.

simonjbeaumont (approved):
> Reverted, the patch touches only CryptoExtras now. I missed the code organisation note in the README about the files without the `_boring` suffix, that is on me.

No worries at all. And thank you for the contribution 🙏

darkdi:
The two workflow runs for this head are sitting in `action_required`, so nothing has reported back and the merge is blocked on missing checks rather than on the patch. They need the approval click.

simonjbeaumont:
No worries. I'll take care of this tomorrow.

PR #469, opened 2026-09-14
Avoid some allocations in the HMAC initialiser

ptoffy:
This improves the HMAC initialiser by removing some useless allocations. Simply using `.init(key:)`, my benchmarks look like this:

Benchmark | Time before | Time after | Mallocs before | Mallocs after
-- | -- | -- | -- | --
SHA256 32-byte key | 579 ns | 145 ns  | 6 | 2
SHA256 64-byte key | 442 ns | 145 ns | 4 | 2
SHA256 128-byte key | 655 ns | 253 ns | 7 | 3
SHA512 32-byte key | 722 ns | 222 ns | 6 | 2
SHA512 64-byte key | 590 ns | 218 ns | 4 | 2
SHA512 128-byte key | 962 ns | 466 ns | 7 | 3

Adding `@inlinable` would give an extras speedup as well but I'm not sure if that's a welcome change.

### Checklist

- [ ] I've run `./scripts/generate_boilerplate_files_with_gyb.sh` and included updated generated files in a commit of this pull request

### Motivation:

We should allocate as little as possible.

### Modifications:

Rewrite the initialiser body around `withUnsafeTemporaryAllocation` and the hasher's `update(unsafePointer:)` methods.

### Result:

Less allocations, more speed


simonjbeaumont:
Thanks @ptoffy. It looks like this initializer has already had some changes in the 5.x branch. Is this patch still relevant against that branch?

Also, as you may already be aware, this one will require a CryptoKit change if we want to take it:

> Files that do not have the _boring or _xkcp suffix are part of the public API of CryptoKit. Changing these requires passing a higher bar, as any change in these files must be accompanied by a change in CryptoKit itself.

-- source: https://github.com/apple/swift-crypto#code-organisation

0xTim:
@simonjbeaumont Why does this require a CryptoKit change? This is an improvement in the open source library that doesn't touch the API or ABI

simonjbeaumont:
> Why does this require a CryptoKit change? This is an improvement in the open source library that doesn't touch the API or ABI

Because the current posture of this package is to mirror the CryptoKit API and behaviour where possible. In order to achieve this, the API _layer_ (not just the API definitions) are shared with the CryptoKit project, and while this patch doesn't change the API, but it does change the first layer of implementation behind the API.

> When building Swift Crypto for use on Linux or Windows, Swift Crypto builds substantially more code. In particular, we build:
> 
> 1. A vendored copy of BoringSSL's libcrypto.
> 2. A vendored copy of XKCP's libXKCP.
> 3. `The common API of Swift Crypto and CryptoKit.`
> 4. The backing implementation of this common API, which calls into BoringSSL or XKCP.
>
> The API code, and some cryptographic primitives which are directly implemented in Swift, `are exactly the same for both Apple CryptoKit and Swift Crypto.` 

-- source: https://github.com/apple/swift-crypto#implementation (`emphasis added`).

ptoffy:
@simonjbeaumont I've tried it out and while the 5.x change is an improvement, this version is still a bit quicker and avoids 2 allocations (this is all p50 btw)

Benchmark | 5.x time | New time | 5.x mallocs | New mallocs
-- | -- | -- | -- | --
SHA256 32-byte key | 343 ns | 145 ns | 4 | 2
SHA256 64-byte key | 378 ns | 145 ns | 4 | 2
SHA256 128-byte key | 475 ns | 253 ns | 5 | 3
SHA512 32-byte key | 435 ns | 222 ns | 4 | 2
SHA512 64-byte key | 420 ns | 218 ns | 4 | 2
SHA512 128-byte key | 679 ns | 466 ns | 5 | 3

simonjbeaumont:
~On the question of the 5.x branch, it looks like that has already reduced allocations for these paths, but not by as much as this PR would have.~ looks like we got to this conclusion at the same time :)

ptoffy:
If this were to trigger a CryptoKit change anyway maybe it'd make sense to sneak in a few `@inlinable`s as well

simonjbeaumont:
@ptoffy let me see if there's scope to land this patch upstream. :)

Sources/Crypto/Message Authentication Codes/HMAC/HMAC.swift
```diff
@@ -65,34 +65,28 @@ public struct HMAC<H: HashFunction>: MACAlgorithm, Sendable {
...
+                } else {
+                    UnsafeMutableRawBufferPointer(padded).copyMemory(from: keyBytes)
+                }
             }
+
+            for i in 0..<blockSize { padded[i] ^= 0x36 }
+            self.innerHasher.update(bufferPointer: UnsafeRawBufferPointer(padded))
+            for i in 0..<blockSize { padded[i] ^= 0x36 ^ 0x5c }
+            self.outerHasher.update(bufferPointer: UnsafeRawBufferPointer(padded))
+            padded.update(repeating: 0)
```
simonjbeaumont, line 88: Would need to take a look at the compiled code to be sure, but I'm concerned this isn't going to provide the same guarantees as the `SecureBytes` it replaces.

I think we should still be able to use the temporary allocation, though, because we have some `zeroize()` functions in `Zeroization.swift` that use `memset_s` (which uses `OPENSSL_cleanse` on Linux).

ptoffy: Now using `.zeroize()`. This shows no measurable performance difference from before

PR #472, opened 2026-09-15
Update to github-workflows soundness v0.0.13

simonjbeaumont:
## Motivation

The 5.x branch updated the soundness GitHub workflow and so the version of swift-format is updated on the 5.x branch CI. We're about to merge 5.x into main but it would be nice to do so without breaking the CI.

## Modifications

Update main to use 0.0.13 soundness swiftlang workflow -- the same version as the 5.x branch.

## Result

Just formatting changes will be required in this patch.


simonjbeaumont:
Let's drop this one and instead do a catchup merge from main into 5.x before merging 5.x back in.

